PR #6, opened 2026-04-27
fix: parse Loki timestamps as Unix nanoseconds, pass limit to Splunk count

### LINKED ISSUE #4: LogsCollector misparses Loki nanosecond timestamps and falls back to time.Now()
### Description

`internal/collector/logs.go` models Loki stream entries as `[timestamp_ns, log_line]`, but `FetchLogs` currently tries to parse the timestamp with `time.RFC3339Nano` and then falls back to `time.Now()` when parsing fails.

That means real Loki timestamps (nanoseconds since epoch) are turned into the current wall-clock time instead of the actual log time.

**Suggested fix**

- Parse `value[0]` as `strconv.ParseInt(..., 10, 64)` and convert with `time.Unix(0, ns).UTC()`
- optionally keep RFC3339Nano as a fallback for compatibility
- skip malformed entries instead of assigning `time.Now()`

### Steps to Reproduce

The current code path is in `internal/collector/logs.go`:
- Loki values are documented as `[timestamp_ns, log_line]`
- parsing currently does:
  - `time.Parse(time.RFC3339Nano, value[0])`
  - else `ts = time.Now()`

The existing success test in `internal/collector/logs_test.go` uses RFC3339 strings, so it does not cover real Loki timestamp format.

### Expected Behaviour

Loki timestamps should be parsed as Unix nanoseconds. If parsing still fails, the entry should be skipped rather than assigned `time.Now()`.

### Actual Behaviour

`internal/collector/logs.go` models Loki stream entries as `[timestamp_ns, log_line]`, but `FetchLogs` currently tries to parse the timestamp with `time.RFC3339Nano` and then falls back to `time.Now()` when parsing fails.

That means real Loki timestamps (nanoseconds since epoch) are turned into the current wall-clock time instead of the actual log time.

### Wachd Version

v0.4.13

### Deployment Method

Docker Compose (local dev)

### Relevant Logs

```shell

```

### LINKED ISSUE #5: SplunkCollector ignores requested limit and always sends count=100
### Description

`internal/collector/splunk.go` accepts a `limit` parameter in both `FetchLogs` and `FetchNotableEvents`, but `runSearch` hardcodes the Splunk `count` form field to 100.

**Suggested fix**

Pass the requested limit through to `runSearch` and use it for the `count` form field, with a fallback default only if limit <= 0.

### Steps to Reproduce

Inspect the current implementation:
- `FetchLogs(..., limit int)` builds the SPL query with `head %d`
- `FetchNotableEvents(..., limit int)` also accepts `limit`
- but `runSearch` always does `form.Set("count", "100")`

### Expected Behaviour

If the caller requests `limit=250`, the Splunk search request should use `count=250` so the API can actually return up to that many results.

### Actual Behaviour

`internal/collector/splunk.go` accepts a `limit` parameter in both `FetchLogs` and `FetchNotableEvents`, but `runSearch` hardcodes the Splunk `count` form field to 100.

### Wachd Version

v0.4.13

### Deployment Method

Docker Compose (local dev)

### Relevant Logs

```shell

```

ntcdeveloper:
## What

Two collector bugs reported in #4 and #5.

### #4 — Loki timestamp parsing (`internal/collector/logs.go`)

Loki returns timestamps as Unix nanoseconds (e.g. `1736942400000000000`), not RFC3339 strings. The old code tried RFC3339 first, always failed, then fell back to `time.Now()` — giving every log entry a fake timestamp equal to when the alert was processed.

Fix: parse as `int64` nanoseconds first, keep RFC3339Nano as fallback for non-standard setups, skip malformed entries entirely.

### #5 — Splunk `count` hardcoded (`internal/collector/splunk.go`)

`FetchLogs` and `FetchNotableEvents` both accepted a `limit` parameter and used it in the SPL `head %d` clause, but `runSearch` always sent `count=100` to the Splunk API regardless. The API enforces its own cap based on `count`, so the SPL limit was silently overridden.

Fix: thread `limit` through `runSearch` into the `count` form field; default to 100 when `limit <= 0`.

## Tests added

- `TestLogsCollector_FetchLogs_NanosecondTimestamp` — real Loki ns format produces correct time
- `TestLogsCollector_FetchLogs_MalformedTimestampSkipped` — bad timestamp skips entry, not time.Now()
- `TestSplunkCollector_FetchLogs_CountPropagated` — limit=250 sends count=250 to Splunk
- `TestSplunkCollector_FetchLogs_CountDefaultsTo100` — limit=0 sends count=100

## Checklist

- [x] `go test ./internal/collector/... ` — all 52 tests pass
- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean

Closes #4
Closes #5

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+# Every file in the repo requires review from the maintainer.
+# Contributors cannot merge their own PRs without this approval.
+* @ntcdeveloper
--- a/internal/collector/logs.go
+++ b/internal/collector/logs.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 )
 
@@ -109,13 +110,16 @@ func (l *LogsCollector) FetchLogs(ctx context.Context, query string, since, unti
 				continue
 			}
 
-			// Parse timestamp (nanoseconds)
+			// Parse timestamp — Loki sends Unix nanoseconds as a string.
+			// Keep RFC3339Nano as a fallback for test doubles and non-standard setups.
+			// Skip the entry entirely if neither format parses — never substitute time.Now().
 			var ts time.Time
-			if nsec, err := time.Parse(time.RFC3339Nano, value[0]); err == nil {
-				ts = nsec
+			if ns, err := strconv.ParseInt(value[0], 10, 64); err == nil {
+				ts = time.Unix(0, ns).UTC()
+			} else if t, err := time.Parse(time.RFC3339Nano, value[0]); err == nil {
+				ts = t
 			} else {
-				// Fallback: parse as unix nanoseconds
-				ts = time.Now()
+				continue
 			}
 
 			logLine := LogLine{
--- a/internal/collector/logs_test.go
+++ b/internal/collector/logs_test.go
@@ -17,6 +17,7 @@ package collector
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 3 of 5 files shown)


ntcdeveloper:
@fatkobra the fix is up — would you mind validating? Both issues are covered in this PR.

fatkobra:
Hey @ntcdeveloper 

Thanks — it looks correct to me for both #4 and #5.

The Loki change addresses the real nanosecond timestamp format instead of falling back to `time.Now()`, and the Splunk change fixes the current `count=100` hardcoding so the requested limit is actually honored.

Optional nice-to-have: a `FetchNotableEvents` count test for symmetry, since #5 affects both callers, although the shared `runSearch` path already makes the main fix look solid. 
Not a blocker from my side, just a nice to have.

ntcdeveloper:
@fatkobra thanks for the thorough reports and the validation — exactly the kind of feedback that makes OS projects better. Really appreciate it.

PR #9, opened 2026-04-28
Fix/oncall userid and layerorder

### LINKED ISSUE #7: GetEscalationChain ignores EscalationLayer.UserID even though fixed-user layers are documented as supported
### Description

`internal/oncall/schedule.go` documents `EscalationLayer` as supporting exactly one of `ScheduleID` or `UserID`, where `UserID` targets a fixed person directly.

However, `Manager.GetEscalationChain(...)` currently only parses and resolves `ScheduleID`. It never checks `layer.UserID`, so fixed-user escalation layers are silently skipped.

**Why this matters?**

This makes a documented escalation mode unusable. A config that intends to notify a fixed person directly will not work even though the type comment says it should.

**Affected code**

- `EscalationLayer` comment documents `UserID`
- `GetEscalationChain` only parses `layer.ScheduleID`

**Suggested fix**

Teach `GetEscalationChain` to:
- support `UserID`
- reject invalid layers where both `ScheduleID` and `UserID` are set
- reject invalid layers where both are empty

I will be more than happy to submit a PR if ok with you.

### Steps to Reproduce

1. Configure an escalation policy whose layer sets `user_id` only, with no `schedule_id`, for example:

   {
     "layers": [
       {
         "user_id": "11111111-1111-1111-1111-111111111111",
         "notify_after_minutes": 0,
         "layer_name": "primary"
       }
     ],
     "repeat_interval_minutes": 0,
     "max_repeats": 0
   }

2. Call `GetEscalationChain(ctx, teamID)` for a team that has:
   - a valid escalation policy using the config above
   - a matching team member with that `user_id`

3. Observe the current behavior:
   - the implementation parses `layer.ScheduleID` unconditionally
   - because `schedule_id` is empty, the layer is skipped
   - if there are no other layers, `GetEscalationChain` returns an error like:
     `escalation policy has no resolvable layers for team ...`

### Expected Behaviour

A layer that sets `user_id` (with no `schedule_id`) should resolve directly to that fixed user and appear in the escalation chain.

### Actual Behaviour

The direct-user layer is ignored because `GetEscalationChain` only handles `schedule_id`.

### Wachd Version

0.4.13

### Deployment Method

Docker Compose (local dev)

### Relevant Logs

```shell

```

### LINKED ISSUE #8: LayerOrder is documented but not enforced in layered on-call resolution
### Description

`internal/oncall/schedule.go` documents `Layer.LayerOrder` as the ordering field ("1 = first notified"), but the layered schedule resolution code does not sort layers before using them.

- `resolveLayered(...)` iterates `rot.Layers` directly
- `resolveLayeredAllLayers(...)` also iterates `rot.Layers` directly
- there is even a comment saying "assume already sorted"

This means behavior depends on JSON array order instead of `LayerOrder`.

For layered schedules, the current on-call user and the per-layer results can be wrong if the layers are stored out of order in the JSON config.

**Suggested fix**

Copy `rot.Layers`, sort them by `LayerOrder`, and iterate the sorted slice in both `resolveLayered(...)` and `resolveLayeredAllLayers(...)`.

Will be happy to submit a PR also.

### Steps to Reproduce

Steps to reproduce

1. Create a layered rotation config where the JSON array order does NOT match `layer_order`, for example:

   {
     "type": "layered",
     "rotation_start": "2025-01-06T00:00:00Z",
     "rotation_interval_hours": 168,
     "layers": [
       {
         "id": "secondary",
         "name": "Secondary",
         "layer_order": 2,
         "time_restriction": { "type": "always" },
         "members": ["22222222-2222-2222-2222-222222222222"]
       },
       {
         "id": "primary",
         "name": "Primary",
         "layer_order": 1,
         "time_restriction": { "type": "always" },
         "members": ["11111111-1111-1111-1111-111111111111"]
       }
     ]
   }

2. Call `resolveLayered(raw, t)` (or any code path that resolves the current on-call user from this rotation) for a time where both layers are active.

3. Observe the current behavior:
   - the code iterates `rot.Layers` directly in array order
   - it returns the first active layer from the JSON slice
   - in the example above, that means "Secondary" wins even though `layer_order=1` is on "Primary"

4. Also call `ResolveAllLayersAt(raw, t)` and observe that results are returned in the raw JSON order, not sorted by `layer_order`.

### Expected Behaviour

Layered schedule resolution should sort by `LayerOrder` before:
- picking the first active layer in `resolveLayered`
- returning all layers from `ResolveAllLayersAt`

### Actual Behaviour

Both functions currently use the JSON slice order instead of sorting by `layer_order`.

### Wachd Version

0.4.13

### Deployment Method

Docker Compose (local dev)

### Relevant Logs

```shell

```

fatkobra:
## Summary
Fixes & Closes #7
Fixes & Closes #8

- support `EscalationLayer.UserID` in `GetEscalationChain`
- reject invalid escalation layers where both `schedule_id` and `user_id` are set
- reject invalid escalation layers where both are empty
- sort layered schedules by `LayerOrder` before resolving
- use sorted layers in both `resolveLayered` and `ResolveAllLayersAt`
- include the maintainer-provided regression tests from `test/oncall-bug-specs`
- add coverage for the both-empty validation case

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Testing

- gofmt -w internal/oncall/schedule.go internal/oncall/schedule_test.go
- go test ./internal/oncall/...
- go test ./...
- go vet ./...
- go build ./...

## Code Review Checklist

## Related Issues

#7 
#8

### DIFF
--- a/internal/oncall/schedule.go
+++ b/internal/oncall/schedule.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"sort"
 	"time"
 
 	"github.com/google/uuid"
@@ -38,14 +39,14 @@ type WeeklyRotation struct {
 
 // DayAssignment maps a weekday name to a user.
 type DayAssignment struct {
-	Day    string    `json:"day"`     // monday…sunday
+	Day    string    `json:"day"` // monday…sunday
 	UserID uuid.UUID `json:"user_id"`
 }
 
 // LayeredRotation is the new format supporting multiple simultaneous on-call
 // layers, time-window restrictions, and index-based rotation.
 type LayeredRotation struct {
-	Type                  string  `json:"type"` // "layered"
+	Type                  string  `json:"type"`                    // "layered"
 	RotationStart         string  `json:"rotation_start"`          // RFC3339 anchor time
 	RotationIntervalHours float64 `json:"rotation_interval_hours"` // e.g. 168 = weekly
 	Layers                []Layer `json:"layers"`
@@ -63,8 +64,8 @@ type Layer struct {
 // TimeRestriction controls when a layer is active.
 type TimeRestriction struct {
 	// Type: "always" | "weekdays" | "weekends" | "custom"
-	Type    string         `json:"type"`
-	Windows []TimeWindow   `json:"windows,omitempty"` // only for type="custom"
+	Type    string       `json:"type"`
+	Windows []TimeWindow `json:"windows,omitempty"` // only for type="custom"
 }
 
 // TimeWindow is a recurring weekly time block (start inclusive, end exclusive).
@@ -131,6 +132,28 @@ func ResolveAllLayersAt(raw []byte, t time.Time) ([]LayerResult, error) {
 	}
 }
 
+// sortedLayers returns a copy of the input ordered by LayerOrder.
+// Positive LayerOrder values are sorted ascending (1 = first notified).
+// Layers with zero or negative LayerOrder are left after explicitly ordered layers.
+func sortedLayers(layers []Layer) []Layer {
+	out := append([]Layer(nil), layers...)
+	sort.SliceStable(out, func(i, j int) bool {
+		li, lj := out[i].LayerOrder, out[j].LayerOrder
+
+		switch {
+		case li <= 0 && lj <= 0:
+			return false
+		case li <= 0:
+			return false
+		case lj <= 0:
+			return true
+		default:
+			return li < lj
+		}
+	})
+	return out
+}
+
 func resolveLayeredAllLayers(raw []byte, t time.Time) ([]LayerResult, error) {
 	var rot LayeredRotation
 	if err := json.Unmarshal(raw, &rot); err != nil {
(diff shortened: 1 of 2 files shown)


ntcdeveloper (approved):
Clean fix on both counts.\n\n**#8 (LayerOrder):** `sortedLayers()` correctly copies the slice before sorting — no mutation of the original. `sort.SliceStable` is the right call here, preserving relative order for equal LayerOrder values. Applied consistently in both `resolveLayered` and `resolveLayeredAllLayers`.\n\n**#7 (UserID):** The `(ScheduleID == "") == (UserID == "")` validation is clean — catches both-empty and both-set in one check. UserID path resolves directly without schedule lookup as intended.\n\nFull suite passes, `go vet` and `go build` clean. ✅ Approved.

ntcdeveloper:
Merged — thank you @fatkobra. Clean implementation, good edge case handling on the LayerOrder zero/negative values, and appreciated the extra `BothIDsEmpty` test. Three PRs in two days — you're a core part of this project now.

PR #11, opened 2026-04-29
fix: store timestamps as UTC in all time.Now() calls

### LINKED ISSUE #10: fix: store timestamps as UTC in all time.Now() calls
## Problem

When running Wachd on a host in a non-UTC timezone (e.g. Norway, UTC+2), `time.Now()` returns local time. When this value is passed to PostgreSQL as a `TIMESTAMPTZ`, the local time is stored without the correct offset — effectively treating Norway local time as UTC.

This causes a 2-hour skew in stored timestamps (`fired_at`, `created_at`, `updated_at`, etc.). The most visible symptom: the worker queries GitHub for commits using `since = incident.FiredAt - 30min`, but `FiredAt` is 2 hours ahead of real UTC, so the window misses recent commits entirely and the AI analysis has no context.

## This does not affect Azure deployments

AKS containers run in UTC, so `time.Now()` and `time.Now().UTC()` produce identical results there. The bug only surfaces on developer machines in non-UTC timezones.

## Proposed fix

Replace all `time.Now()` with `time.Now().UTC()` in `internal/store/` so timestamps written to the DB are always explicitly UTC regardless of host timezone.

Affected files:
- `internal/store/incidents.go`
- `internal/store/local_users.go`
- `internal/store/team_config.go`
- `internal/store/system_config.go`
- `internal/store/notification_rules.go`

## Question for reviewers

Is `time.Now().UTC()` the right fix, or should this be handled at the `pgx` connection level (e.g. `TimeZone=UTC` in the connection string)? Both approaches should work — want a second opinion before opening a PR.

ntcdeveloper:
Closes #10

## What this fixes

On developer machines in non-UTC timezones (e.g. Norway UTC+2), `time.Now()` returns local time. Stored in a `TIMESTAMP` column and later treated as UTC, this caused a 2-hour skew — the GitHub commit window used for AI analysis missed recent commits entirely.

## Changes

**Go side (`time.Now().UTC()` in all store writes):**
- `internal/store/incidents.go` — `CreateIncident`, `UpdateIncidentStatus`, `AcknowledgeIncident`, `GetOpenIncidentsForEscalation`, `IncrementEscalationStep`
- `internal/store/local_users.go` — `RecordLocalLogin`
- `internal/store/system_config.go` — `GetSystemConfig` default, `UpsertSystemConfig`, `SeedSystemConfig`
- `internal/store/team_config.go` — `UpsertTeamConfig`
- `internal/store/notification_rules.go` — `UpsertUserNotificationRule`, `QueuePendingNotification`

**SQL side (`TimeZone=UTC` pgx runtime param in `NewDB`):**
- `internal/store/db.go` — sets `config.ConnConfig.RuntimeParams["TimeZone"] = "UTC"` so `NOW()` and `DEFAULT NOW()` in SQL also write UTC

This implements exactly the two-part fix @fatkobra recommended in #10.

## No production impact

AKS/EKS containers run in UTC — `time.Now()` and `time.Now().UTC()` produce identical results there. This only affects developer machines in non-UTC timezones.

@fatkobra — would appreciate your review given you diagnosed the SQL-side gap.

### DIFF
--- a/internal/store/db.go
+++ b/internal/store/db.go
@@ -51,6 +51,12 @@ func NewDB(databaseURL string) (*DB, error) {
 		config.ConnConfig.Password = pw
 	}
 
+	// Force the PostgreSQL session timezone to UTC so that SQL-side NOW() and
+	// DEFAULT NOW() write UTC regardless of the host machine's local timezone.
+	// This complements time.Now().UTC() on the Go side — both must be UTC to
+	// avoid skew on developer machines in non-UTC timezones (e.g. Norway UTC+2).
+	config.ConnConfig.RuntimeParams["TimeZone"] = "UTC"
+
 	// Set connection pool settings
 	config.MaxConns = 20
 	config.MinConns = 2
--- a/internal/store/incidents.go
+++ b/internal/store/incidents.go
@@ -35,7 +35,7 @@ func (db *DB) CreateIncident(ctx context.Context, incident *Incident) error {
 		)
 	`
 
-	now := time.Now()
+	now := time.Now().UTC()
 	incident.ID = uuid.New()
 	incident.CreatedAt = now
 	incident.UpdatedAt = now
@@ -165,7 +165,7 @@ func (db *DB) UpdateIncidentStatus(ctx context.Context, teamID, incidentID uuid.
 		SET status = $1, updated_at = $2
 		WHERE id = $3 AND team_id = $4
 	`
-	now := time.Now()
+	now := time.Now().UTC()
 	_, err := db.pool.Exec(ctx, query, status, now, incidentID, teamID)
 	if err != nil {
 		return fmt.Errorf("failed to update incident status: %w", err)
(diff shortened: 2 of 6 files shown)


fatkobra:
This is the right short-term fix.

From the current store/schema setup, you switched the Go-side writes to UTC, but you also made sure SQL-side `NOW()` / `DEFAULT NOW()` paths stay aligned. Setting the pgx session `TimeZone=UTC` in `NewDB` closes that gap nicely.


That feels like the correct operational fix for the skew reported on non-UTC developer machines. 

At a later stage it would be great to think about a future follow-up to migrate the core store columns from `TIMESTAMP` to `TIMESTAMPTZ`, this would still make sense long-term.

No blockers from my side — very good job i catching this, and also very good job with the fix.

ntcdeveloper:
Thanks @fatkobra — really appreciate the thorough review and catching the SQL-side gap. The TIMESTAMPTZ migration is noted, will track that as a follow-up.

Issue #18, opened 2026-04-30
security: webhook rate limiter X-Forwarded-For trust is still spoofable in Kubernetes deployments


ntcdeveloper:
## Summary

cmd/server/main.go:579 trusts `X-Forwarded-For` only when `RemoteAddr` is a private/loopback address. In a typical Kubernetes/ingress deployment, `RemoteAddr` is **always** a private IP (the ingress node), so this check effectively trusts `X-Forwarded-For` unconditionally in production — the same vulnerability it was meant to fix.

If the edge proxy (nginx-ingress, Traefik, etc.) is configured to append rather than overwrite the `X-Forwarded-For` header, an attacker can rotate the header value on every request and bypass per-IP webhook rate limiting.

## Fix options

**Code-side (preferred):** Add a `TRUSTED_PROXY_CIDRS` config value (e.g. in `values.yaml`). Only trust `X-Forwarded-For` when `RemoteAddr` falls within those configured CIDRs. Default to empty (trust nothing) so new installs are safe by default.

**Deployment-side (required regardless):** Document that operators must configure their ingress to **overwrite** `X-Forwarded-For` with the real client IP, not append to it. For nginx-ingress this is `use-forwarded-headers: "false"` or `compute-full-forwarded-for: "false"`.

Both mitigations are needed — code alone can't protect against a misconfigured ingress.

## References

- Partial fix in this PR: cmd/server/main.go:571

ntcdeveloper:
@fatkobra — same here. This is the X-Forwarded-For / trusted-proxy CIDR follow-up. The current fix is only safe if the ingress overwrites XFF, which most k8s setups don't do by default. Let us know if you want to take a stab at this or have thoughts on the right approach.

fatkobra:
Here I think the `TRUSTED_PROXY_CIDRS` direction is the right approach, and I’d default it to empty (trust nothing) exactly as you suggested.

The main implementation detail I’d push for is this:

- only consider `X-Forwarded-For` if the immediate `RemoteAddr` is in a configured trusted proxy CIDR
- when it is trusted, derive the client IP from the XFF chain **right-to-left**, not by taking the first value

Why right-to-left:
if an ingress appends instead of overwriting, the left-most element can still contain attacker-controlled junk. Walking from the right side of the chain is much safer for proxy setups and multi-hop trusted proxies.

I’d also suggest:
- parse/canonicalize IPs with `net/netip` or equivalent, not raw string slicing
[...]

Issue #17, opened 2026-04-30
security: SSRF protection does not cover transport — DNS rebinding and redirect attacks still possible


ntcdeveloper:
## Summary

`validate.EndpointURL` validates hostnames at config-save time (cmd/server/main.go:1354), but the actual outbound HTTP clients resolve DNS independently at connection time. This leaves two attack vectors open:

**DNS rebinding:** a hostname resolves to a public IP during validation, then rebinds to `127.0.0.1` or a cloud metadata IP (`169.254.169.254`) by the time the real connection is made.

**Open redirects:** if an outbound HTTP client follows redirects, the final redirect target is never validated against private ranges.

Affected outbound callers (non-exhaustive):
- `internal/collector/logs.go` — Loki/Datadog endpoint
- `internal/notify/slack.go` — Slack webhook URL
- `internal/collector/metrics.go` — Prometheus endpoint

## Fix

Apply a custom `http.Transport` to all outbound clients with:
1. **`DialContext`** that resolves the hostname and checks every returned IP against `validate.isPrivateIP` before opening the TCP connection — TOCTOU-safe because validation and connection use the same resolved IP.
2. **`CheckRedirect`** that calls `validate.EndpointURL` on every redirect target, or disables redirects entirely for endpoints that don't require them.

## References

- Original validation-time fix: #13
- Affected files: `internal/collector/logs.go:88`, `internal/notify/slack.go:167`, `internal/collector/metrics.go`

ntcdeveloper:
@fatkobra — flagging you on this one too if you have bandwidth. This is the transport-level SSRF fix (custom DialContext + CheckRedirect) that the PR summary mentioned as deferred. Would value your input on the approach before we implement.

fatkobra:
I think the overall direction here is right, but I recommend centralizing it as one reusable outbound transport/helper rather than fixing callers one by one.

A couple of implementation details I’d treat as important:

1. The custom `DialContext` needs to validate the resolved IP **and then actually dial that validated IP**, not just resolve/validate and then pass the original hostname back to the default dial path. Otherwise the DNS rebinding / TOCTOU window is still there.

2. The transport should preserve the original hostname for HTTP Host / TLS verification semantics while dialing the validated IP underneath.

3. I would apply the same hardened transport everywhere outbound HTTP is constructed, not just the examples in the issue. In current code that includes at least:
   - `internal/collector/logs.go`
[...]

Issue #16, opened 2026-04-30
security: on-call override creation does not verify target user belongs to the team


ntcdeveloper:
## Severity
Medium

## Description
\`handleCreateOverride\` at \`cmd/server/main.go:1861\` accepts any \`user_id\` in the request body and inserts the override without checking that the user is a member of the team. Any team responder or admin who knows (or guesses) an identity UUID outside the team can route on-call notifications to that person.

## Location
- \`cmd/server/main.go:1911\` — override inserted without membership check

## Fix
Before inserting the override, load the team's member list via \`GetTeamMembers(teamID)\` and verify the requested \`user_id\` is present. Return \`400 Bad Request\` if not.

ntcdeveloper:
Fixed in PR #19 (merged): override creation now verifies the target user belongs to the team; cross-team user returns 422.

Issue #15, opened 2026-04-30
security: webhook rate limiter trusts client-supplied X-Forwarded-For header


ntcdeveloper:
## Severity
Medium

## Description
\`cmd/server/main.go:573\` uses the first value from the \`X-Forwarded-For\` header unconditionally to choose the rate limiter bucket. If the server is reachable directly (no proxy), an attacker can bypass rate limiting entirely by rotating this header on every request.

## Location
- \`cmd/server/main.go:573\`

## Fix
Only trust \`X-Forwarded-For\` when the request originates from a known trusted proxy IP. Otherwise fall back to \`r.RemoteAddr\` exclusively. A simple approach: make trusted proxy CIDRs configurable (e.g. via env var), and only apply the header extraction when \`RemoteAddr\` is in that set.

ntcdeveloper:
Fixed in PR #19 (merged): XFF header only trusted when RemoteAddr is a private/loopback address (in-cluster proxy).

PR #19, opened 2026-04-30
fix: security hardening — auth, SSRF, rate limit, override access

ntcdeveloper:
## Summary

Six security fixes from an external audit, plus integration test repairs.

**Fixes in this PR:**

| Issue | Fix |
|---|---|
| #12 — Config endpoints exposed webhook secret to non-admins | GET/PUT config now require admin role; webhook_secret removed from response |
| #13 — SSRF protection skipped DNS resolution | EndpointURL resolves DNS at validation time, checks all returned IPs against private/reserved ranges |
| #14 — No webhook body size limit | MaxBytesReader(1 MB) added before io.ReadAll |
| #15 — X-Forwarded-For trusted unconditionally | XFF only trusted when RemoteAddr is private/loopback (in-cluster proxy) |
| #16 — Override target not verified as team member | GetTeamMembers check before CreateOverride; cross-team user → 422 |
| (unissued) — Phone update not scoped to team | Same membership check before UpdateMemberPhone |

**Also fixed:**
- Viewers could create on-call overrides — `requireTeamResponder` added, viewers now get 403
- Viewer denial regression test added

**Integration test repairs (tests were broken before this PR):**
- `webhookLimiter` and `cfg` were nil in test server — nil pointer panics on webhook calls
- Config test used `prometheus.example.com` / `loki.example.com` which now fail DNS validation — removed fake endpoints
- Webhook security test expected 403 for wrong-secret cases but handler returns 401 — fixed expectations

**Deferred (separate issues):**
- #17 — SSRF DNS rebinding via transport (custom DialContext + CheckRedirect)
- #18 — XFF still spoofable in k8s with misconfigured ingress (configurable trusted-proxy CIDRs)

## Test plan

- [ ] `go build ./...` — clean
- [ ] `go vet ./...` — clean
- [ ] `go test ./...` — all pass
- [ ] `go test -tags integration ./cmd/server/` — requires local Postgres + Redis
- [ ] Verify config GET no longer includes `webhook_secret` in response
- [ ] Verify viewer role gets 403 on override creation
- [ ] Verify cross-team user_id returns 422 on override creation and phone update

/cc @fatkobra — would appreciate a second look, especially on the requireTeamResponder logic and the integration test repairs

### DIFF
--- a/internal/validate/url.go
+++ b/internal/validate/url.go
@@ -15,14 +15,60 @@
 package validate
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/url"
 	"strings"
+	"time"
 )
 
+// privateNets lists CIDR ranges that must never be the target of an outbound request.
+var privateNets = func() []*net.IPNet {
+	cidrs := []string{
+		"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16",
+		"127.0.0.0/8", "169.254.0.0/16",
+		"::1/128", "fc00::/7", "fe80::/10",
+		"100.64.0.0/10", // Carrier-grade NAT
+		"224.0.0.0/4",   // Multicast
+		"240.0.0.0/4",   // Reserved
+	}
+	nets := make([]*net.IPNet, 0, len(cidrs))
+	for _, cidr := range cidrs {
+		_, n, err := net.ParseCIDR(cidr)
+		if err == nil {
+			nets = append(nets, n)
+		}
+	}
+	return nets
+}()
+
+// resolveHost is the DNS lookup function used by EndpointURL. Tests may
+// override this variable to avoid real network calls.
+var resolveHost = func(ctx context.Context, host string) ([]string, error) {
+	return net.DefaultResolver.LookupHost(ctx, host)
+}
+
+// isPrivateIP returns true if ip falls within any private or reserved range.
+func isPrivateIP(ip net.IP) bool {
+	if v4 := ip.To4(); v4 != nil {
+		ip = v4
+	}
+	if ip.Equal(net.IPv4zero) || ip.Equal(net.IPv6unspecified) {
+		return true
+	}
+	for _, n := range privateNets {
+		if n.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
 // EndpointURL rejects URLs that could be used for SSRF attacks.
-// Only public http/https URLs are allowed — no private IP ranges, no localhost.
+// Only public http/https URLs are allowed. For hostnames, DNS is resolved at
+// validation time and every returned address is checked against
+// private/loopback/link-local/reserved ranges.
 func EndpointURL(rawURL string) error {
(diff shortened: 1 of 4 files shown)


fatkobra:
Thanks — I took another look at the actual implementation, and I think there is one blocker before I’d call this fully good to merge.

**Blocker: the SSRF validator still does not cover all of the “private/reserved” space the PR summary claims to block.**

In `internal/validate/url.go`, the new `privateNets` list covers RFC1918, loopback, link-local, CGNAT, multicast, and 240/4, but it still omits several special-purpose / reserved blocks such as:
- `192.0.0.0/24` (IANA special-purpose / protocol assignments)
- `198.18.0.0/15` (benchmarking)
- the RFC 5737 documentation ranges (`192.0.2.0/24`, `198.51.100.0/24`, `203.0.113.0/24`)
- the IPv6 documentation prefix `2001:db8::/32`

Right now the tests actually codify at least one of those as allowed: `TestEndpointURL_Allowed` stubs DNS to `203.0.113.1`, which is in TEST-NET-3 / RFC 5737, so the current implementation is still treating at least one reserved block as public. The PR summary says DNS validation “checks all returned IPs against private/reserved ranges”, but the implementation is narrower than that.

So I think the choices are:
1. **tighten the implementation/tests** to include the missing special-purpose ranges, or
2. **narrow the wording/scope** so it explicitly says this blocks RFC1918/loopback/link-local/etc. rather than “all private/reserved ranges”.

Everything else in the PR looks good to me, but I would treat that mismatch as a real blocker for the SSRF hardening claim.

ntcdeveloper:
Good catch on both counts @fatkobra.

Fixed in 24c968e:
- Added the missing ranges to `privateNets`: `192.0.0.0/24` (IANA special-purpose), `198.18.0.0/15` (benchmarking), `192.0.2.0/24`, `198.51.100.0/24`, `203.0.113.0/24` (RFC 5737 TEST-NET-1/2/3), `2001:db8::/32` (IPv6 documentation)
- Swapped the test DNS stub from `203.0.113.1` (now correctly blocked) to `8.8.8.8` which is genuinely public and outside all blocked ranges

All four `url_test.go` tests still pass.

internal/validate/url.go
```diff
@@ -26,12 +26,24 @@ import (
...
+		// RFC 1918 — private
 		"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16",
+		// Loopback and link-local
 		"127.0.0.0/8", "169.254.0.0/16",
+		// IPv6 loopback, ULA, link-local
 		"::1/128", "fc00::/7", "fe80::/10",
-		"100.64.0.0/10", // Carrier-grade NAT
-		"224.0.0.0/4",   // Multicast
-		"240.0.0.0/4",   // Reserved
+		// Carrier-grade NAT (RFC 6598)
```
fatkobra, line 35: Looks very good now.

fatkobra:
No blockers now.

PR #22, opened 2026-04-30
fix: return 404 when incident does not exist

### LINKED ISSUE #20: GET /api/v1/teams/{teamId}/incidents/{incidentId} dereferences nil for missing incidents instead of returning 404
### Description

Description

`internal/store/incidents.go` returns `(nil, nil)` from `GetIncident(...)` when a row is not found, but `cmd/server/main.go`'s `handleGetIncident` only checks `err` and then immediately calls `incident.ToResponse()`.

That means a valid request for a non-existent incident can dereference a nil incident instead of returning a clean 404. The handler also currently maps any store error to 404, which can hide real DB failures.

Affected code

- `internal/store/incidents.go` — `GetIncident` returns `nil, nil` on no rows
- `cmd/server/main.go` — `handleGetIncident` calls `incident.ToResponse()` without checking `incident == nil`

Suggested fix

In `handleGetIncident`:
- if `err != nil`, return 500
- if `incident == nil`, return 404
- only call `ToResponse()` when `incident != nil`

I will submit a PR with proposed fix and test.

### Steps to Reproduce

1. Start Wachd with a valid team and a logged-in user who has access to that team.
2. Send:
   `GET /api/v1/teams/{teamId}/incidents/{random-uuid-that-does-not-exist}`
3. Observe the current behavior:
   - the store returns `(nil, nil)` for the missing incident
   - the handler does not check for `nil`
   - it proceeds to `incident.ToResponse()`

### Expected Behaviour

A missing incident ID should return HTTP 404 cleanly.

### Actual Behaviour

The handler can dereference a nil incident and does not distinguish “not found” from real store failures.

### Wachd Version

0.4.14

### Deployment Method

Docker Compose (local dev)

### Relevant Logs

```shell

```

fatkobra:
## Summary

Closes #20 

Fixes missing incident retrieval so the server returns 404 instead of dereferencing a nil incident.

- distinguish store errors from "not found" in `handleGetIncident`
- return 500 on real DB errors
- return 404 when `GetIncident` returns nil
- add an e2e regression test for GET on a non-existent incident ID

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Testing

- gofmt -w cmd/server/main.go cmd/server/e2e_integration_test.go
- go test ./...
- go vet ./...
- go build ./...

## Code Review Checklist

### DIFF
--- a/cmd/server/e2e_integration_test.go
+++ b/cmd/server/e2e_integration_test.go
@@ -659,3 +659,25 @@ func TestE2E_WebhookSecurity(t *testing.T) {
 		})
 	}
 }
+
+// TestE2E_GetIncident_NotFoundReturns404 verifies that requesting a
+// non-existent incident ID returns HTTP 404 instead of dereferencing a nil
+// incident.
+func TestE2E_GetIncident_NotFoundReturns404(t *testing.T) {
+	env := newE2EEnv(t)
+
+	missingID := uuid.New()
+
+	req := httptest.NewRequest(
+		"GET",
+		fmt.Sprintf("/api/v1/teams/%s/incidents/%s", env.kongTeam.ID, missingID),
+		nil,
+	)
+	req.AddCookie(env.kongCookie)
+
+	resp := env.do(req)
+
+	if resp.Code != http.StatusNotFound {
+		t.Fatalf("get missing incident: got %d, want 404\nbody: %s", resp.Code, resp.Body)
+	}
+}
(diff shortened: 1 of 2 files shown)


ntcdeveloper (approved):
Fix is correct — 500 on store error, 404 on nil incident, ToResponse() only called when non-nil. Test covers the case. The formatting/alignment changes are harmless but unrelated to the bug — consider keeping fixes separate from style cleanup in future PRs. Merging.

PR #24, opened 2026-05-06
fix(security): add SSRF-safe HTTP transport

fatkobra:
## Summary

Part of #17.

This adds a centralized `internal/safehttp` package for outbound HTTP clients.

The helper protects against the two transport-level SSRF gaps described in #17:

- DNS rebinding between config validation and actual connection time
- unvalidated redirects to private/internal targets

## What changed

- Added `safehttp.NewTransport()`.
  - Resolves the request hostname inside `DialContext`.
  - Validates every resolved IP.
  - Dials the already-validated IP directly instead of handing the hostname back to `net.Dial`.
- Added `safehttp.NewRoundTripper()`.
  - Validates the initial request URL with the existing `validate.EndpointURL`.
  - Uses the safe transport for the actual network connection.
- Added client helpers:
  - `CollectorClient()` validates redirect targets before following them.
  - `WebhookClient()` disables redirects for webhook-style POSTs.
- Added tests for:
  - private DNS results
  - mixed public/private DNS results
  - dialing the validated IP instead of the original hostname
  - private URL rejection before network dial
  - redirect validation
  - webhook redirect disabling

## Why

`validate.EndpointURL` protects config-save-time validation, but it does not protect the actual connection if DNS changes later.

The important part is avoiding the TOCTOU pattern:

1. resolve hostname;
2. validate the resolved IP;
3. accidentally dial the hostname again.

This helper validates and then dials the same resolved IP.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore / dependency update

## Testing

```bash
gofmt -w internal/safehttp/safehttp.go internal/safehttp/safehttp_test.go
go test ./internal/safehttp
go test ./...
make test
go build ./...
go vet ./...
```

- [x] Unit tests pass (`make test`)
- [x] Build passes (`go build ./...`)
- [x] Static analysis passes (`go vet ./...`)
- [x] Tested manually against a running instance

## Code Review Checklist

- [x] All database queries that access team data use `WHERE team_id = $N`
- [x] No hardcoded secrets, API keys, or credentials
- [x] Error handling returns errors — no panics in production paths
- [x] New API endpoints validate input (UUIDs, required fields)
- [x] PII sanitizer is called before any analysis backend call
- [x] Logs include context (incident ID, team ID) where relevant

## Related Issues

Part of #17

PR #28, opened 2026-05-06
fix(security): use safe HTTP clients for outbound calls

### LINKED ISSUE #17: security: SSRF protection does not cover transport — DNS rebinding and redirect attacks still possible

fatkobra:
## Summary

Fixes #17.

This wires the centralized `internal/safehttp` helper into the outbound HTTP callers identified in the issue.

## What changed

- Loki logs collector now uses `safehttp.CollectorClient`.
- Dynatrace collector now uses `safehttp.CollectorClient`.
- Prometheus metrics collector now passes `safehttp.NewRoundTripper()` into the Prometheus API client.
- Slack notifier now uses `safehttp.WebhookClient`.
- Collector redirects are allowed only when every redirect target passes URL validation.
- Slack webhook POSTs do not follow redirects.
- Tests were updated so local `httptest` servers use injected local test clients instead of the production SSRF-blocking client.

## Why

Config-time URL validation is not enough to prevent SSRF.

Without transport-level protection, a hostname can validate as public at save time and then resolve to a private IP at connection time. Redirects can also move an otherwise valid outbound request to a private or metadata-service target.

The safe transport resolves and validates IPs inside `DialContext`, then dials the already-validated IP directly.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation

## Testing

```bash
gofmt -w \
  internal/collector/logs.go \
  internal/collector/dynatrace.go \
  internal/collector/metrics.go \
  internal/notify/slack.go \
  internal/collector/logs_test.go \
  internal/collector/dynatrace_test.go \
  internal/collector/metrics_test.go \
  internal/notify/slack_email_test.go
go test ./internal/safehttp ./internal/collector ./internal/notify
go test ./...
make test
go build ./...
go vet ./..
```

## Code Review Checklist

## Related Issues

Closes #17

### DIFF
--- a/internal/collector/dynatrace.go
+++ b/internal/collector/dynatrace.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/wachd/wachd/internal/safehttp"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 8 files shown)


ntcdeveloper (approved):
## Code Review

**✅ CODE APPROVED**

Pattern is consistent across all four callers — public constructor wires in the safe client, private constructor accepts injection for tests, nil fallback in all private constructors as a safety net.

Client types are correct:
- Loki → `CollectorClient` (GET, redirects validated) ✅
- Dynatrace → `CollectorClient` ✅
- Prometheus → `NewRoundTripper()` injected into the Prometheus API config — correct for that library ✅
- Slack → `WebhookClient` (POST, redirects blocked entirely) ✅

Test pattern is right — existing `httptest.NewServer` tests bypass the safe transport via `srv.Client()` or `http.DefaultTransport`. Without this, those tests would fail because the safe transport blocks loopback IPs.

`TestSlackNotifierDoesNotFollowRedirects` is the best new test — actually fires a redirect and asserts the target is never reached.

**✅ SECURITY APPROVED**

All four outbound callers identified in #17 are now protected. DNS rebinding and redirect attacks are blocked at the transport level. Closes the gap left after #24.

Minor note: import ordering in `slack.go` and `slack_email_test.go` is slightly off (stdlib should come before third-party). Not a blocker.

PR #30, opened 2026-05-07
fix(security): trust forwarded IPs only from configured proxies

### LINKED ISSUE #18: security: webhook rate limiter X-Forwarded-For trust is still spoofable in Kubernetes deployments

fatkobra:
## Summary

Fixes #18.

This changes webhook rate-limiting client IP extraction so `X-Forwarded-For` is trusted only when the immediate peer is explicitly configured as a trusted proxy.

## What changed

- Added `TRUSTED_PROXY_CIDRS`.
  - Empty by default.
  - Empty means trust no `X-Forwarded-For` headers.
- Added trusted proxy CIDR parsing using `net/netip`.
- Added safe client IP extraction for the webhook rate limiter:
  - use `RemoteAddr` by default
  - only consider `X-Forwarded-For` when `RemoteAddr` is in `TRUSTED_PROXY_CIDRS`
  - parse the full XFF chain
  - walk the chain right-to-left
  - ignore malformed XFF and fall back to `RemoteAddr`
- Added tests for:
  - empty/default trust config
  - invalid CIDR config
  - ignoring XFF from untrusted peers
  - ignoring XFF when no proxies are trusted
  - right-to-left XFF parsing through trusted proxy hops
  - malformed XFF fallback
  - IPv6 proxy/client parsing

## Why

The old logic trusted `X-Forwarded-For` whenever the direct peer was private or loopback. In Kubernetes, the direct peer is commonly an ingress or in-cluster proxy, so this effectively made `X-Forwarded-For` attacker-controlled in production.

This PR makes proxy trust explicit.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation

## Testing

```bash
gofmt -w cmd/server/main.go cmd/server/proxy_test.go
go test ./cmd/server
go test ./...
make test
go build ./...
go vet ./...
```

## Code Review Checklist

## Related Issues

Closes #18 but there are 1 more PR after this one.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -27,6 +27,7 @@ import (
 	"math/big"
 	"net"
 	"net/http"
+	"net/netip"
 	"os"
 	"os/signal"
 	"strconv"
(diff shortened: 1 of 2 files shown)


ntcdeveloper:
Good fix — the algorithm is correct and the test coverage is solid.

One thing worth noting for the follow-up PR: `TRUSTED_PROXY_CIDRS` needs to be added to `values.yaml` and the env var list in the README so operators know to configure it. Without it, deployments behind an ingress controller will rate-limit per-ingress-IP rather than per-client-IP after this change — which is safe but different from the current behaviour.

Also worth a one-liner in the changelog or release notes flagging this as a behaviour change for existing deployments.

Otherwise looks good — merging.

fatkobra:
Thanks — before I open the follow-up PR, I want to confirm the intended Helm shape.

Current `main` already has:

```yaml
ingress:
  trustedProxyCIDRs: ""
```
and the ConfigMap already passes that through as TRUSTED_PROXY_CIDRS.

So for PR2, should I:

keep the existing ingress.trustedProxyCIDRs value and add the missing README env-var docs + changelog behavior-change note; or
move/add the value under security.trustedProxyCIDRs as originally sketched in the issue?

My preference is option 1 to avoid introducing a second Helm value for the same setting, unless you prefer the security.* location for chart organization.

Issue #36, opened 2026-05-19
feature: smarter incident similarity matching


fatkobra:
## Problem

When an alert fires, Wachd should be able to tell whether the same or a very similar incident happened before.

Right now, the most valuable user-facing feature from the incident knowledge graph would be: "we have seen this before, and here is how it was resolved."

## Proposed solution

Use the incident knowledge graph and embeddings to compare new incidents against previous incidents.

Similarity could consider:
- root cause text
- log patterns
- affected service
- metric anomalies
- recent commits/deploys
- previous resolutions

When a new incident is created, Wachd could show the most similar past incidents with:
- similarity score
- previous root cause
- previous resolution
- affected service
- incident timestamp

## Why this matters

This would make Wachd more useful during real incidents because teams could quickly reuse previous knowledge instead of starting from zero.

This connects directly to issue #35, because the graph can store the incident history and similarity edges.

### Alternatives Considered

Several alternatives were considered:

1. Exact keyword or log search only  
This would be simpler to build, but it would miss incidents that are related even when the wording, log message, or alert title is slightly different.

2. Manual tagging by users  
Teams could manually tag incidents with labels like `database`, `deployment`, or `redis`, but this adds extra work during or after incidents. It also depends on people remembering to tag things consistently.

3. Match only by affected service  
Wachd could show previous incidents from the same service, but that would be too broad. Many incidents can affect the same service while having completely different root causes.

4. Let the AI compare incidents without storing structured graph data  
This would be easier at first, but it would be harder to explain why two incidents were considered similar. Using the incident knowledge graph makes the matching more transparent and reusable.

5. Do nothing and rely on team memory  
This is what many teams already do today, but it does not scale well. As incidents pile up, it becomes harder for engineers to remember previous fixes.

### Feature Area

AI analysis / incident knowledge graph

fatkobra:
I opened a PR for this here: #40 

The PR adds the first backend implementation for smarter incident similarity matching on top of the graph foundation from #35/#39.

What it includes:

- a Postgres-backed graph store implementation
- `FindSimilar` support for incident nodes
- weighted similarity scoring using:
  - alert title
  - root cause
  - log patterns
  - affected service
  - metric anomalies
  - deploy/commit context
  - previous resolution
- similarity scores
- human-readable match reasons
- unit tests for the similarity scoring logic
- optional Postgres integration test coverage

[...]

PR #39, opened 2026-05-19
feat: incident knowledge graph foundation

### LINKED ISSUE #35: design: incident knowledge graph — node types, edge types, AI write-back
## Background

We're building an incident knowledge graph into Wachd. Instead of analyzing each alert in isolation, the system builds a persistent graph of incidents, services, commits, root causes, and resolutions — so the AI has memory and can surface "this happened before, here's how it was resolved."

Two architectural decisions are already made:

- **Storage:** Postgres adjacency table + pgvector for similarity matching, behind a `GraphStore` interface (`internal/graph/store.go`). Same swappable pattern as `AIBackend` and `Notifier` — ship `PostgresGraphStore` first, swap later without touching callers.
- **UI:** Basic graph visualization ships in v1 — enough to see the feature. Stronger visualization in a later iteration.

Three decisions need community input before we write code.

---

## Question 1 — Node types

Candidates so far:

| Node type | What it represents | Source |
|---|---|---|
| `incident` | The alert event — service, severity, timestamp, AI root cause | Wachd DB |
| `service` | The affected service — name, repo, team | team_data_sources |
| `commit` | Recent commits at alert time | collector/git.go |
| `metric_anomaly` | The metric that crossed threshold | collector/metrics.go |
| `log_pattern` | Recurring error pattern in the log window | collector/logs.go |
| `root_cause` | AI-derived cause — text + embedding vector | ai/ output |
| `resolution` | How it was resolved — rollback, hotfix, escalate | incident resolve action |

**Open questions:**
- Should `service` be an explicit registry node (teams configure services) or just a label on `incident` nodes? Explicit nodes enable the dependency graph (issue #27) but require teams to configure them.
- Should we store the extracted `log_pattern` (e.g. `"connection refused on port 5432"`) or a pointer back to the log source? Pattern is better for similarity; pointer is better for drill-down.

---

## Question 2 — Edge types

Candidates so far:

| Edge type | From → To | Answers |
|---|---|---|
| `caused_by` | incident → root_cause | "why did this fire?" |
| `similar_to` | incident → incident | "has this happened before?" |
| `depends_on` | service → service | "what else could be affected?" |
| `triggered_after` | incident → commit | "did a deploy cause this?" |
| `resolved_by` | incident → resolution | "how was it fixed last time?" |
| `affects` | incident → service | "which service owns this incident?" |
| `preceded_by` | incident → metric_anomaly | "what was the leading signal?" |

Note: `similar_to` edges are computed automatically — pgvector finds nearby root cause embeddings above a threshold and creates the edge. All others are built at collection time.

**Open question:** Are there edge types missing that would make the graph more useful for your team's incident workflow?

---

## Question 3 — When does AI write back to the graph?

Three options:

**A — After resolution only**
Incident closes → AI writes root cause + resolution to graph. Simple. Graph only helps future incidents, not the active one.

**B — Incrementally during the incident**
Alert fires → AI writes draft root cause immediately → enriches as investigation progresses → resolution commits the final edges. Most powerful but complex — partial writes, concurrent updates.

**C — Two-phase: draft on fire, commit on resolve**
Alert fires → AI writes a `pending` subgraph → on resolution, promoted to permanent nodes/edges. Draft nodes excluded from similarity matching. Clean discard path for false alarms.

**Open question:** Which model fits how your team actually works incidents?

---

Tagging @fatkobra — you've seen the codebase most closely, your input here would be valuable before we start building.

ntcdeveloper:
## What this adds

The foundation layer for the incident knowledge graph (closes #35 foundation step; unblocks #36).

**`internal/graph/store.go`** — `Store` interface with:
- `Node` type — `incident | deployment | service | alert`
- `Edge` type — `caused_by | affects | similar_to | triggered`
- `Graph` type — subgraph container for API responses
- Five interface methods: `UpsertNode`, `UpsertEdge`, `GetSubgraph`, `FindSimilar`, `DeleteNode`
- Every method is team-scoped (`teamID uuid.UUID` as required param)

**`internal/store/schema.sql`** — two new tables:
- `graph_nodes` — stores nodes with JSONB properties, partial unique index on `(team_id, type, external_id)`
- `graph_edges` — directed edges with `UNIQUE(from_id, to_id, type)` for safe upserts, `ON DELETE CASCADE` on both FK sides

## What comes next (#36)

The Postgres implementation of `graph.Store` can now be built on top of this interface. Suggested starting point: `internal/graph/postgres.go` implementing each method with parameterized queries against the two new tables.

## Checklist

- [x] `go build ./internal/graph/...` clean
- [x] `go vet ./internal/graph/...` clean
- [x] All interface methods are team-scoped
- [x] Schema uses `IF NOT EXISTS` — idempotent on startup
- [x] `ON DELETE CASCADE` on all FKs — no orphaned graph data

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,6 @@ Wachd receives alerts from your monitoring tools (Grafana, Datadog, Prometheus) 
 [![Go Version](https://img.shields.io/badge/go-1.24%2B-00ADD8.svg)](go.mod)
 [![Go Report Card](https://goreportcard.com/badge/github.com/wachd/wachd)](https://goreportcard.com/report/github.com/wachd/wachd)
 [![Discord](https://img.shields.io/badge/Discord-join-5865F2?logo=discord&logoColor=white)](https://discord.gg/XAgXgnqxa)
-[![GitHub Stars](https://img.shields.io/github/stars/wachd/wachd?style=social)](https://github.com/wachd/wachd)
-
-If Wachd is useful to you, a ⭐ on GitHub helps other engineers find it.
 
 ![Wachd demo — on-call check, alert fire, AI root cause analysis](docs/demo.gif)
 
--- a/internal/graph/store.go
+++ b/internal/graph/store.go
@@ -1,0 +1,101 @@
+// Copyright 2025 NTC Dev
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package graph
+
+import (
+	"context"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// NodeType defines what kind of entity a graph node represents.
+type NodeType string
+
+const (
+	NodeTypeIncident   NodeType = "incident"
+	NodeTypeDeployment NodeType = "deployment"
+	NodeTypeService    NodeType = "service"
+	NodeTypeAlert      NodeType = "alert"
+)
+
+// EdgeType defines the relationship between two graph nodes.
+type EdgeType string
+
+const (
+	// EdgeTypeCausedBy links an incident to the deployment or incident that caused it.
+	EdgeTypeCausedBy EdgeType = "caused_by"
+	// EdgeTypeAffects links an incident to the service it impacted.
+	EdgeTypeAffects EdgeType = "affects"
+	// EdgeTypeSimilarTo links two incidents with overlapping symptoms or context.
+	EdgeTypeSimilarTo EdgeType = "similar_to"
+	// EdgeTypeTriggered links an alert rule to the incident it created.
+	EdgeTypeTriggered EdgeType = "triggered"
+)
+
+// Node is a vertex in the incident knowledge graph.
+type Node struct {
+	ID         uuid.UUID `json:"id"`
+	TeamID     uuid.UUID `json:"team_id"`
+	Type       NodeType  `json:"type"`
+	Label      string    `json:"label"`
+	ExternalID *string   `json:"external_id,omitempty"` // incidents.id, commit hash, service name, etc.
+	Properties []byte    `json:"properties,omitempty"`  // JSONB — type-specific metadata
+	CreatedAt  time.Time `json:"created_at"`
+	UpdatedAt  time.Time `json:"updated_at"`
+}
+
+// Edge is a directed relationship between two nodes.
+type Edge struct {
+	ID         uuid.UUID `json:"id"`
+	TeamID     uuid.UUID `json:"team_id"`
+	FromNodeID uuid.UUID `json:"from_node_id"`
+	ToNodeID   uuid.UUID `json:"to_node_id"`
+	Type       EdgeType  `json:"type"`
+	Weight     float64   `json:"weight"`
+	Properties []byte    `json:"properties,omitempty"` // JSONB — relationship metadata
+	CreatedAt  time.Time `json:"created_at"`
+}
+
+// Graph is a subgraph result returned by GetSubgraph.
+type Graph struct {
+	Nodes []*Node `json:"nodes"`
+	Edges []*Edge `json:"edges"`
+}
+
+// Store is the interface for reading and writing the incident knowledge graph.
+// All methods are scoped to a single team — no cross-team access is possible.
+type Store interface {
+	// UpsertNode creates or updates a graph node.
+	// Within a team, (type, external_id) uniquely identifies a node when external_id is set.
+	// If external_id is nil, a new node is always created.
+	UpsertNode(ctx context.Context, teamID uuid.UUID, n *Node) (*Node, error)
+
+	// UpsertEdge creates or updates a directed edge between two nodes.
+	// (from_node_id, to_node_id, type) is unique — calling again updates weight/properties.
+	UpsertEdge(ctx context.Context, teamID uuid.UUID, e *Edge) (*Edge, error)
+
+	// GetSubgraph returns a root node and all connected nodes and edges up to the
+	// given traversal depth. Depth 1 returns only the root and its immediate neighbours.
+	GetSubgraph(ctx context.Context, teamID uuid.UUID, rootNodeID uuid.UUID, depth int) (*Graph, error)
+
+	// FindSimilar returns up to limit nodes of the same type whose label is
+	// similar to the label of the given node. Used to surface past incidents
+	// with matching context before the AI analysis runs.
+	FindSimilar(ctx context.Context, teamID uuid.UUID, nodeID uuid.UUID, limit int) ([]*Node, error)
+
+	// DeleteNode removes a node and all edges connected to it.
+	DeleteNode(ctx context.Context, teamID uuid.UUID, nodeID uuid.UUID) error
+}
--- a/internal/graph/store.go
+++ b/internal/graph/store.go
@@ -16,6 +16,7 @@ package graph
 
 import (
 	"context"
+	"encoding/json"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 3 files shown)


ntcdeveloper:
@fatkobra — this is the foundation for issue #35 / the incident knowledge graph we discussed.

Once this merges, issue #36 is yours if you want it. Starting point would be `internal/graph/postgres.go` — implement each method in the `Store` interface against the two new tables (`graph_nodes`, `graph_edges`). Happy to review whenever you're ready.

fatkobra:
Thanks for putting this foundation together. Overall this looks like the right direction for #35 and it gives #36 a clear place to start. I like the separate `internal/graph` package, the explicit `Store` interface, the team-scoped method signatures, and the pending/permanent status idea for two-phase write-back.

A few things I think are worth tightening before this becomes the base for the Postgres implementation:

### 1. Team isolation should be enforced at the DB edge level

`graph_edges` has its own `team_id`, and `from_id` / `to_id` reference `graph_nodes(id)`, but I do not think the schema currently guarantees that both endpoint nodes belong to the same team as the edge.

That means a bug in the future Postgres implementation could accidentally create an edge like:

- edge.team_id = team A
- from_id = node from team A
- to_id = node from team B

The interface is team-scoped, which is good, but I think this should also be impossible at the database level.

One way to enforce it:

```sql
ALTER TABLE graph_nodes
ADD CONSTRAINT graph_nodes_team_id_id_unique UNIQUE (team_id, id);

CREATE TABLE IF NOT EXISTS graph_edges (
    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    team_id UUID NOT NULL REFERENCES teams(id) ON DELETE CASCADE,
    from_id UUID NOT NULL,
    to_id UUID NOT NULL,
    type VARCHAR(50) NOT NULL,
    weight FLOAT NOT NULL DEFAULT 1.0,
    properties JSONB,
    created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
    updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),

    FOREIGN KEY (team_id, from_id)
        REFERENCES graph_nodes(team_id, id)
        ON DELETE CASCADE,

    FOREIGN KEY (team_id, to_id)
        REFERENCES graph_nodes(team_id, id)
        ON DELETE CASCADE,

    UNIQUE(team_id, from_id, to_id, type)
);

**Comment by fatkobra** [comment]
I looked through the graph foundation and I think the direction makes sense.

I had a few questions/suggestions. These are not blockers from my side, but I think they are worth considering before this becomes the base for the Postgres implementation.

### 2. Properties JSON response

I noticed `Node.Properties` and `Edge.Properties` are currently `[]byte`.

Would it be better to use `json.RawMessage` here?

My understanding is that plain `[]byte` can be serialized as base64 in JSON responses, which may make the API output harder to read. Since these fields represent JSONB metadata, `json.RawMessage` might keep the API response closer to the original JSON object.

For example, instead of:

    Properties []byte `json:"properties,omitempty"`

Maybe use:

    Properties json.RawMessage `json:"properties,omitempty"`

This is not a blocker, but I think it may make the API response easier to work with later.

---

### 3. Similarity results may need score/reason

`FindSimilar` currently returns only `[]*Node`.

For the similarity matching feature, I think the caller may need more than just the matching node. It would be useful to know why something matched.

For example, the result could include:

- matched node
- similarity score
- short reason/evidence

Maybe something like:

    type SimilarNode struct {
        Node   *Node   `json:"node"`
        Score  float64 `json:"score"`
        Reason string  `json:"reason,omitempty"`
    }

Then `FindSimilar` could return:

    FindSimilar(ctx context.Context, teamID uuid.UUID, nodeID uuid.UUID, limit int) ([]*SimilarNode, error)

This would make the feature easier to explain in the UI later, for example:

> This incident looks similar because it affected the same service and had a similar log pattern.

---

### 4. Should edges also have pending/permanent status?

The PR adds `status` to `graph_nodes`, which makes sense for two-phase AI write-back.

I am wondering whether `graph_edges` should also have a status.

For example, during an active incident, the AI may create a draft relationship like:

- incident A `caused_by` deployment B
- incident A `similar_to` incident C

If that relationship is only a draft, should the edge also be marked `pending` until the incident is resolved?

Possible options:

1. Add `status` to `graph_edges` too.
2. Guarantee that pending/draft edges always touch a pending node.
3. Document clearly that pending edges must not appear in similarity or traversal results.

Option 1 seems safest to me, but I am open to your preferred design.

---

### 5. Pending nodes and active incident viewing

The comments say pending nodes must never appear in `FindSimilar` or `GetSubgraph` results.

That makes sense for avoiding draft AI data from polluting future similarity searches.

One thing I am not fully sure about: during an active incident, would the UI need to show the pending incident graph?

Maybe the intended rule is:

> Pending nodes should be excluded from similarity/memory lookup, but the requested root node may be pending when viewing the active incident itself.

If that is the intent, it may be worth documenting in the interface comment so the Postgres implementation handles active incidents correctly.

---

### 6. Database checks for valid node/edge types

The Go code defines valid node types and edge types, but the database currently stores them as plain `VARCHAR`.

Would it make sense to add `CHECK` constraints so invalid values cannot be inserted accidentally?

For example, for node status:

    status VARCHAR(20) NOT NULL DEFAULT 'pending'
        CHECK (status IN ('pending', 'permanent'))

For node type:

    type VARCHAR(50) NOT NULL
        CHECK (type IN ('incident', 'deployment', 'service', 'alert'))

For edge type:

    type VARCHAR(50) NOT NULL
        CHECK (type IN ('caused_by', 'affects', 'similar_to', 'triggered'))

This would make the database enforce the same rules as the Go code.

---

### 7. Future graph types from issue #35

This PR starts with these node types:

- incident
- deployment
- service
- alert

And these edge types:

- caused_by
- affects
- similar_to
- triggered

That is a good foundation.

I am wondering how the other concepts from issue #35 will be represented later, such as:

- root cause
- resolution
- log pattern
- metric anomaly
- commit
- triggered_after
- resolved_by
- preceded_by
- depends_on

Should those become first-class node/edge types later, or should they be stored inside `properties`?

I think either approach could work, but it would help to document the intended direction because similarity matching, timelines, and postmortems will probably depend on those concepts.

---

### 8. `updated_at` for graph edges

`graph_nodes` has both `created_at` and `updated_at`, but `graph_edges` only has `created_at`.

Since `UpsertEdge` can update an existing edge's weight/properties, it may be useful for edges to also have:

    updated_at TIMESTAMPTZ NOT NULL DEFAULT now()

That would make it easier to know when a relationship was last changed.

**Comment by ntcdeveloper** [comment]
All 8 points addressed in 8c1e93e. Going through them:

**1 — Composite FK for team isolation**
Added `UNIQUE(team_id, id)` to `graph_nodes`. `graph_edges` now uses composite FKs:
```sql
FOREIGN KEY (team_id, from_id) REFERENCES graph_nodes(team_id, id) ON DELETE CASCADE,
FOREIGN KEY (team_id, to_id)   REFERENCES graph_nodes(team_id, id) ON DELETE CASCADE,
```
Also updated the `UNIQUE` on edges to `(team_id, from_id, to_id, type)`. Cross-team edges are now rejected at the DB level.

**2 — `json.RawMessage` for Properties**
Changed `[]byte` → `json.RawMessage` on both `Node.Properties` and `Edge.Properties`. The base64 serialization issue is gone.

**3 — `FindSimilar` return type**
Added `SimilarNode`:
```go
type SimilarNode struct {
    Node   *Node   `json:"node"`
    Score  float64 `json:"score"`
    Reason string  `json:"reason,omitempty"`
}
```
`FindSimilar` now returns `[]*SimilarNode`.

**4 — Edge status**
Added `EdgeStatus` type with `pending` / `permanent` constants. `Edge` now has a `Status EdgeStatus` field. Schema has `status VARCHAR(20) NOT NULL DEFAULT 'pending' CHECK (status IN ('pending', 'permanent'))` on `graph_edges`.

**5 — Pending root node visibility**
Clarified in the `GetSubgraph` doc comment:
> The root node is returned regardless of its status — callers displaying the active incident's own graph are not blocked by the pending state. Only neighbour nodes reached via edge traversal are filtered to `status = 'permanent'`.

**6 — CHECK constraints**
Both tables have CHECK on `type` and `status`. Typos now fail at write time with a clear error instead of silently storing unqueryable rows.

**7 — Future node/edge types**
Added to both `NodeType` and `EdgeType` doc comments, and to the schema block comment:
> New types go as first-class constants in `store.go` and as CHECK values in the schema — never as free-form strings in `properties`.

**8 — `updated_at` on `graph_edges`**
Added `updated_at TIMESTAMPTZ NOT NULL DEFAULT now()` to `graph_edges`. `Edge` struct has `UpdatedAt time.Time`.

PR #40, opened 2026-05-21
feat(graph): add smarter incident similarity matching

### LINKED ISSUE #36: feature: smarter incident similarity matching

fatkobra:
## Summary

Closes #36.

This PR adds the first backend implementation of smarter incident similarity matching on top of the incident knowledge graph foundation from #35/#39.

It adds:

- `PostgresStore`, a Postgres-backed implementation of the `graph.Store` interface
- metadata-aware `FindSimilar` support for incident nodes
- weighted similarity scoring across:
  - alert title
  - root cause text
  - log patterns
  - affected service
  - metric anomalies
  - commits/deployments
  - previous resolution
- similarity score and human-readable reason output through the existing `SimilarNode` type
- unit tests for the scoring logic
- optional integration tests for the Postgres-backed graph store

## Why

When an alert fires, Wachd should be able to say:

> This looks like a previous incident. Here is how similar it is and why it matched.

This makes the graph immediately useful to responders and creates the backend path for showing previous root causes, resolutions, affected services, and timestamps in the UI later.

## Notes

This implementation intentionally does not add a pgvector dependency yet. The current schema does not have an embedding/vector column, so this PR adds a deterministic metadata-aware matcher first. The scoring logic is isolated so it can be replaced or extended with pgvector/embedding similarity later without changing the `FindSimilar` API.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation

## Tests

```bash
go test ./internal/graph
go test ./...
```

## Code Review Checklist

### DIFF
--- a/internal/graph/postgres_store.go
+++ b/internal/graph/postgres_store.go
@@ -26,10 +26,7 @@ import (
 	"github.com/jackc/pgx/v5/pgxpool"
 )
 
-const (
-	defaultSimilarityCandidateLimit = 200
-	defaultMinimumSimilarityScore   = 0.12
-)
+const defaultMinimumSimilarityScore = 0.12
 
 // ErrNodeNotFound is returned when a graph operation cannot find the requested
 // node inside the requested team boundary.
@@ -40,8 +37,8 @@ var ErrNodeNotFound = errors.New("graph node not found")
 // It stores graph nodes and edges in the adjacency tables created by
 // internal/store/schema.sql. All operations are scoped by team_id.
 type PostgresStore struct {
-	pool               *pgxpool.Pool
-	candidateLimit     int
+	pool *pgxpool.Pool
+
 	minSimilarityScore float64
 }
 
@@ -51,7 +48,6 @@ var _ Store = (*PostgresStore)(nil)
 func NewPostgresStore(pool *pgxpool.Pool) *PostgresStore {
 	return &PostgresStore{
 		pool:               pool,
-		candidateLimit:     defaultSimilarityCandidateLimit,
 		minSimilarityScore: defaultMinimumSimilarityScore,
 	}
 }
@@ -349,11 +345,10 @@ func (s *PostgresStore) FindSimilar(ctx context.Context, teamID uuid.UUID, nodeI
 		return nil, err
 	}
 
-	candidateLimit := s.candidateLimit
-	if expandedLimit := limit * 10; expandedLimit > candidateLimit {
-		candidateLimit = expandedLimit
-	}
-
+	// Score every permanent same-team candidate instead of applying a recency
+	// cutoff. This keeps older but highly similar incidents discoverable. A
+	// future embeddings/pgvector implementation can replace this with indexed
+	// nearest-neighbour retrieval without changing the FindSimilar API.
 	rows, err := s.pool.Query(ctx, `
 		SELECT id, team_id, type, status, label, external_id, properties, created_at, updated_at
 		FROM graph_nodes
@@ -361,9 +356,7 @@ func (s *PostgresStore) FindSimilar(ctx context.Context, teamID uuid.UUID, nodeI
 		  AND type = $2
 		  AND status = 'permanent'
 		  AND id <> $3
-		ORDER BY updated_at DESC
-		LIMIT $4
-	`, teamID, source.Type, source.ID, candidateLimit)
+	`, teamID, source.Type, source.ID)
 	if err != nil {
 		return nil, fmt.Errorf("query similar candidates: %w", err)
 	}
--- a/internal/graph/postgres_store.go
+++ b/internal/graph/postgres_store.go
@@ -213,74 +213,93 @@ func (s *PostgresStore) GetSubgraph(ctx context.Context, teamID uuid.UUID, rootN
 
 	nodesByID := map[uuid.UUID]*Node{root.ID: root}
 	edgesByID := map[uuid.UUID]*Edge{}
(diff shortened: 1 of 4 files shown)


ntcdeveloper:
Reviewed `postgres_store.go`, `similarity.go`, and the tests. This is solid work.

**What's done well:**

The `PromoteNode` transaction is the hard part of the two-phase design — promoting the node first, then promoting edges only when both endpoints are permanent via the two EXISTS subqueries. That runs inside the same transaction so the EXISTS checks see the just-promoted status. Non-obvious and correct.

The demotion guard in both `UpsertNode` and `UpsertEdge` (`CASE WHEN status = 'permanent' THEN status ELSE EXCLUDED.status END`) means re-upserting a node during a subsequent incident never silently unwinds the two-phase guarantee. Good defensive choice.

`normalizeKey` stripping all non-alphanumeric handles `root_cause`, `rootCause`, and `rootcause` resolving to the same key without any explicit case mapping. The stop word list including `service`, `alert`, `error`, `failure` is domain-aware and prevents high-frequency incident words from inflating token scores.

`availableWeight` renormalizing on missing signals means a node with no `root_cause` field isn't artificially penalized — the remaining signals share the full weight. Right call.

**Tests:**

The three tests are behavioral, not structural — they test what matters. `TestScoreNodeSimilarityUsesIssue36Signals` would catch a broken scorer, a broken reason builder, or a weight misconfiguration all at once. `TestScoreNodeSimilarityFallsBackToIncidentTitle` covers the degraded path where a node has no properties yet — important for freshly created nodes before AI analysis runs. `TestScoreNodeSimilarityDoesNotLetServiceAloneDominate` directly validates the weight distribution — if the service weight were accidentally inflated, that test breaks. Good level of abstraction too: written against `scoreNodeSimilarity` directly so they survive internal refactors without changes.

**One known limitation worth a comment:**

`GetSubgraph` runs one query per frontier node per depth level — O(frontier × depth) round trips. For sparse incident graphs at depth 1–2 this is fine. A recursive CTE would collapse it to one query at higher depths. Not blocking for v1 but worth a `// TODO` so it's documented.

**One minor observation:**

`FindSimilar` selects candidates `ORDER BY updated_at DESC LIMIT 200`. This is a recency window — a very old but highly similar incident past position 200 won't be found. The default limit is generous and this is the right tradeoff before pgvector lands, just worth knowing the edge case exists.

✅ Approved. Ready to merge.

fatkobra:
Thanks for the review and approval.

I will work on these today, rather do it now, then later. No rush to merge.

ntcdeveloper:
Sounds good — take your time. Once the two comments are in, this is ready to go.

fatkobra:
Thanks again for the review.

I agreed with the `GetSubgraph` observation and decided to implement it now instead of leaving it as a TODO.

I pushed a follow-up commit that replaces the per-frontier query loop with a recursive CTE, so `GetSubgraph` now performs traversal in one database query instead of one query per frontier node per depth level.

I also added an integration test covering:

- depth 1 traversal
- depth 2 traversal
- pending root node is allowed
- pending neighbour nodes are excluded
- pending/draft graph data does not leak into the returned subgraph

I will submit another commit for the `FindSimilar` 200-candidate recency window.

fatkobra:
I pushed one more follow-up commit for the `FindSimilar` candidate-window observation.

Instead of leaving the 200-candidate recency window in place, `FindSimilar` now scores every permanent same-team candidate of the same node type. That means a very old but highly similar incident can still be found.

I also added an integration test that creates:

- one old highly similar incident
- 205 newer unrelated incidents
- one new source incident

The test verifies that the old similar incident is still returned as the best match.

I still left full embeddings/pgvector retrieval for a future PR because that needs a larger schema and ingestion design, but this removes the current recency blind spot without changing the public `FindSimilar` API.

ntcdeveloper:
Both fixes are better than what was asked for. Merging.

PR #48, opened 2026-05-22
fix(store): migrate legacy timestamps to timestamptz

fatkobra:
## Summary

Closes #25.

This PR migrates legacy PostgreSQL `TIMESTAMP` columns to `TIMESTAMPTZ` across the store schema.

It updates `internal/store/schema.sql` so:

- fresh databases create timestamp columns as `TIMESTAMPTZ`
- existing databases migrate old `timestamp without time zone` columns to `TIMESTAMPTZ`
- existing timestamp values are reinterpreted as UTC using `AT TIME ZONE 'UTC'`
- the migration is idempotent and only alters columns that are still `timestamp without time zone`

## Why

The Go code already writes UTC timestamps and the DB connection forces the session timezone to UTC. This PR brings the database schema in line with that behavior.

## Implementation notes

Instead of running unconditional `ALTER TABLE ... TYPE TIMESTAMPTZ` statements on every startup, this PR adds a guarded migration block that checks `information_schema.columns`.

That means repeat calls to `Migrate()` do not keep rewriting already-migrated `TIMESTAMPTZ` columns.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore / dependency update

## Testing

Added schema-level tests that verify:

- the schema no longer contains bare `TIMESTAMP` column types
- the issue #25 migration block includes every expected legacy timestamp column
- the migration uses `AT TIME ZONE 'UTC'`
- the migration only targets `timestamp without time zone` columns

Commands run:

```bash
go test ./internal/store -run TestSchema
go test ./internal/store
go test ./...
make test
go build ./...
go vet ./...
```
- [x] Unit tests pass (`make test`)
- [x] Build passes (`go build ./...`)
- [x] Static analysis passes (`go vet ./...`)
- [x] Tested manually against a running instance

## Code Review Checklist

- [x] All database queries that access team data use `WHERE team_id = $N`
- [x] No hardcoded secrets, API keys, or credentials
- [x] Error handling returns errors — no panics in production paths
- [x] New API endpoints validate input (UUIDs, required fields)
- [x] PII sanitizer is called before any analysis backend call
- [x] Logs include context (incident ID, team ID) where relevant

Issue #25, opened 2026-05-06
improvement: migrate TIMESTAMP columns to TIMESTAMPTZ across all tables


ntcdeveloper:
## Background

PR #11 fixed the Go code to store UTC values correctly using `time.Now().UTC()`. This is a follow-up schema improvement.

## What

Convert all `TIMESTAMP` (no timezone) columns to `TIMESTAMPTZ` (with timezone) across all tables in `internal/store/schema.sql`.

Tables affected: `teams`, `users`, `incidents`, `schedules`, `sso_identities`, `sessions`, `password_policy`, `local_users`, `local_groups`, `local_group_members`, `group_mappings`, `sso_providers`, `api_tokens`, `schedule_overrides`, `escalation_policies`, `team_config`, `system_config`.

Newer tables (`user_notification_rules`, `pending_notifications`) already use `TIMESTAMPTZ` — this brings the rest of the schema in line.

## Migration approach

Add `ALTER TABLE ... ALTER COLUMN ... TYPE TIMESTAMPTZ USING column AT TIME ZONE 'UTC'` statements to schema.sql. Since the Go code already writes UTC, existing data is safe to reinterpret as UTC.

All ALTER statements must be idempotent — safe to run on existing databases.

## Notes

- This is a schema hygiene improvement, not a bug fix — the system works correctly today
- Needs code review before merge
- Run against a test DB before merging to confirm no table rewrite issues on larger datasets

fatkobra:
I opened a PR for this here: #48 

The PR updates `internal/store/schema.sql` so fresh databases create legacy timestamp columns as `TIMESTAMPTZ`, and existing databases migrate old `timestamp without time zone` columns using:

`USING column AT TIME ZONE 'UTC'`

I also made the migration guarded through `information_schema.columns`, so it only alters columns that are still `timestamp without time zone`. That keeps the migration idempotent and avoids repeatedly rewriting already-migrated columns on every startup.

Added schema tests to verify:

- no bare `TIMESTAMP` column types remain
- every issue #25 legacy timestamp column is listed in the migration
- the migration uses `AT TIME ZONE 'UTC'`
- the migration only targets `timestamp without time zone` columns

PR #49, opened 2026-05-22
improvement: migrate store timestamps to TIMESTAMPTZ

ADITYA-CODE-SOURCE:
fix#25

Why
- PR #11 standardized application-side timestamp writes to UTC and set the PostgreSQL session timezone to UTC.
- The store schema still defined many core time columns as TIMESTAMP (no timezone), which is ambiguous outside the app.

What changed
- Convert legacy TIMESTAMP columns to TIMESTAMPTZ in `internal/store/schema.sql` for fresh installs.
- Add an idempotent migration block that converts existing `timestamp without time zone` columns to TIMESTAMPTZ using `AT TIME ZONE 'UTC'`.
- Add a regression test to ensure the migration runs cleanly twice.

Verification
- `go test -v ./...` (DB-backed store tests skip if Postgres is not running)

### DIFF
--- a/internal/store/schema.sql
+++ b/internal/store/schema.sql
@@ -6,8 +6,8 @@ CREATE TABLE IF NOT EXISTS teams (
     id             UUID         PRIMARY KEY DEFAULT gen_random_uuid(),
     name           VARCHAR(255) NOT NULL,
     webhook_secret VARCHAR(255) NOT NULL UNIQUE,
-    created_at     TIMESTAMP    NOT NULL DEFAULT NOW(),
-    updated_at     TIMESTAMP    NOT NULL DEFAULT NOW()
+    created_at     TIMESTAMPTZ  NOT NULL DEFAULT NOW(),
+    updated_at     TIMESTAMPTZ  NOT NULL DEFAULT NOW()
 );
 
 CREATE INDEX IF NOT EXISTS idx_teams_webhook_secret ON teams(webhook_secret);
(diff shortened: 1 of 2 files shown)


fatkobra:
Thanks for adding this.

Since #48 has already been merged, I do not think this PR is needed as another implementation of issue #25 anymore. The actual schema migration and schema-level tests are already in `main`.

That said, I do think the new DB-backed regression test is useful. It complements #48 by exercising the migration against a real legacy table:

- create a legacy `password_policy.updated_at TIMESTAMP`
- insert a UTC value
- run `Migrate()`
- run `Migrate()` again to prove idempotency
- verify the column becomes `timestamp with time zone`
- verify the stored value is preserved

So my suggestion would be to keep this PR, but reframe it as a test-only follow-up.

Requested cleanup:

1. Rename the PR to something like:

   `test(store): add timestamptz migration regression test`

2. Update the PR body so it says this complements #48 instead of presenting itself as the main fix for #25.

3. Revert the `schema.sql` whitespace/no-op change if possible. The useful part of this PR is the new test in `store_test.go`, and keeping `schema.sql` unchanged would make the diff cleaner.

One small note: GitHub shows the workflow completed with no jobs, so this DB-backed test may not have actually run in CI. It would be good to confirm locally with:

`go test ./internal/store`

or, if the DB-backed tests require Postgres:

`TEST_DATABASE_URL=<postgres-url> go test ./internal/store`

Overall: I think the regression test is a good addition, but the PR should be cleaned up and described as a follow-up test PR rather than the main issue #25 migration.

What do you guys think @ADITYA-CODE-SOURCE @ntcdeveloper   ?

But otherwise, very grateful for your contribution. Keep it up.

ntcdeveloper (approved):
✅ Approved.

The schema.sql change (trailing newline removal) is harmless. The real value here is the DB-backed regression test — it exercises the migration against a live Postgres instance, which the static schema tests in #48 cannot do.

Test coverage is correct:
- Legacy table created with bare TIMESTAMP to simulate a pre-migration database
- Migrate() called twice — idempotency confirmed against real DB, not just SQL inspection
- Column type verified via information_schema.columns
- Stored value verified with got.Equal(legacyTime) — data preservation confirmed
- Cleanup properly drops and remigrates the table

Complements #48. Merging.

Issue #41, opened 2026-05-22
feat(graph): wire graph store into worker — write incidents to graph on resolution


ntcdeveloper:
## Context

`internal/graph` ships a complete `PostgresStore` with `UpsertNode`, `PromoteNode`, and `FindSimilar`. The worker processes and resolves incidents but never calls any of these. The graph is always empty, so `FindSimilar` returns nothing.

## What to implement

In the worker, after an incident is marked resolved:

1. Call `graph.UpsertNode` with the incident data mapped to a `Node`:
   - `Type`: `"incident"`
   - `Label`: alert title / name
   - `Properties`: JSON blob with `root_cause`, `log_pattern`, `service`, `metric_anomaly`, `deployment`, `resolution`
2. Call `graph.PromoteNode` to mark it permanent — only permanent nodes are searched by `FindSimilar`
3. Load the team's graph config from `team_graph_config` (`enabled`, `min_similarity_score`) before writing — skip if disabled

The `PostgresStore` must be initialised in `cmd/worker/main.go` and injected into the incident processing pipeline.

## Testing requirements

Every new function must have both unit and integration tests — this is a non-negotiable project rule.

**Unit tests** (`cmd/worker/` or `internal/`):
- Node mapping: incident fields → `Node.Properties` JSON (correct keys, no PII leakage)
- Skip graph write when team config has `enabled = false`
- Handles `PromoteNode` error without crashing the worker (fail safe — alert routing must still complete)

**Integration tests** (against real Postgres):
- Resolved incident appears as a permanent node in the graph
- Resolving the same incident twice does not duplicate the node (upsert behaviour)
- Node is scoped to the correct `team_id` — query with a different team returns nothing

## Acceptance criteria

- [ ] Resolved incident is written to the graph as a permanent node
- [ ] Worker does not crash or drop alerts if graph write fails
- [ ] All unit and integration tests pass: `go test ./...`
- [ ] No cross-team data leakage (every DB query filters by `team_id`)
- [ ] Sanitiser runs on incident data before it touches the graph (PII must not be stored)

ntcdeveloper:
Closed by PR #50 (merged). Worker now writes resolved incidents to the graph as permanent nodes on resolution.

PR #50, opened 2026-05-23
feat(graph): write resolved incidents to graph in worker

ADITYA-CODE-SOURCE:
Addresses #41

This wires the incident similarity graph end-to-end by writing resolved incidents into the graph store.

What changed
- Added `team_graph_config` table (enabled + min_similarity_score) and store accessors.
- Enqueued an `incident_resolved` worker job from the incident resolve API handler.
- Worker initializes `graph.PostgresStore` and, on resolved jobs:
  - builds a sanitized incident node (label + properties)
  - upserts by external_id (incident ID)
  - promotes node to permanent (fail-safe on errors)
- Added unit tests for node mapping + config gating + fail-safe behavior.
- Added DB-backed integration test (skips if Postgres is unavailable).

Verification
- `go test ./...`
- Note: the Postgres-backed worker integration test requires a running DB (uses `TEST_DATABASE_URL` or defaults to localhost:5432).

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -1255,6 +1255,9 @@ func (s *Server) handleResolveIncident(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "Failed to resolve incident", http.StatusInternalServerError)
 		return
 	}
+	if err := s.queue.EnqueueIncidentResolved(r.Context(), incidentID, teamID); err != nil {
+		log.Printf("Failed to enqueue resolved incident job: %v", err)
+	}
 
 	log.Printf("✓ Incident %s resolved", incident.Title)
 
--- a/cmd/worker/graph.go
+++ b/cmd/worker/graph.go
@@ -28,6 +28,11 @@ import (
 	"github.com/wachd/wachd/internal/store"
 )
 
+const (
+	maxGraphLogLines = 3
+	maxGraphCommits  = 2
+)
+
 type teamGraphConfigReader interface {
 	GetTeamGraphConfig(ctx context.Context, teamID uuid.UUID) (*store.TeamGraphConfig, error)
 }
@@ -54,6 +59,8 @@ func persistResolvedIncidentNode(ctx context.Context, cfgStore teamGraphConfigRe
 		log.Printf("warn: load team graph config for %s: %v", teamID, err)
 		return nil
 	}
+	// min_similarity_score is intentionally unused on the write path; it applies
+	// when FindSimilar reads back permanent nodes.
 	if cfg != nil && !cfg.Enabled {
 		return nil
 	}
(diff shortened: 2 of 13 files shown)


ntcdeveloper (commented):
Good implementation — write path is solid, fail-safe pattern is consistent, and the integration test running the job twice to verify upsert idempotency is a nice touch.

A few discussion points before we merge:

**Double sanitisation on analysis fields**
`loadIncidentAnalysis` sanitises `RootCause`, `SuggestedAction`, and `KeyFindings` before returning, then `buildResolvedIncidentNode` passes those same values through `sanitiseGraphValue` again. Functionally correct (running the sanitiser twice on clean input is a no-op), but worth picking one canonical place so the intent is clear. Happy to leave it as-is if the double-pass was deliberate for defence-in-depth — just want to make sure it was a conscious choice.

**`min_similarity_score` loaded but not used in this path**
`persistResolvedIncidentNode` loads the full `TeamGraphConfig` and checks `Enabled`, but `MinSimilarityScore` is never referenced. That field belongs to the read path (`FindSimilar`), not the write path, so this is correct — but without a comment a future reader will wonder if something was forgotten. A short inline note would close that loop.

**Magic numbers in the summarize helpers**
`summarizeLogPattern` caps at `3` lines and `summarizeDeployment` caps at `2` commits. Both are hardcoded literals. Named constants (`maxGraphLogLines`, `maxGraphCommits`) would make it obvious these are intentional limits and easy to tune later.

None of these are blockers — just flagging for discussion.

ADITYA-CODE-SOURCE:
> Good implementation — write path is solid, fail-safe pattern is consistent, and the integration test running the job twice to verify upsert idempotency is a nice touch.
> 
> A few discussion points before we merge:
> 
> **Double sanitisation on analysis fields** `loadIncidentAnalysis` sanitises `RootCause`, `SuggestedAction`, and `KeyFindings` before returning, then `buildResolvedIncidentNode` passes those same values through `sanitiseGraphValue` again. Functionally correct (running the sanitiser twice on clean input is a no-op), but worth picking one canonical place so the intent is clear. Happy to leave it as-is if the double-pass was deliberate for defence-in-depth — just want to make sure it was a conscious choice.
> 
> **`min_similarity_score` loaded but not used in this path** `persistResolvedIncidentNode` loads the full `TeamGraphConfig` and checks `Enabled`, but `MinSimilarityScore` is never referenced. That field belongs to the read path (`FindSimilar`), not the write path, so this is correct — but without a comment a future reader will wonder if something was forgotten. A short inline note would close that loop.
> 
> **Magic numbers in the summarize helpers** `summarizeLogPattern` caps at `3` lines and `summarizeDeployment` caps at `2` commits. Both are hardcoded literals. Named constants (`maxGraphLogLines`, `maxGraphCommits`) would make it obvious these are intentional limits and easy to tune later.
> 
> None of these are blockers — just flagging for discussion.

@ntcdeveloper  Thanks for the review.
- Double sanitisation: not deliberate—happy to simplify. I’ll keep sanitisation in one canonical place (likely sanitiseGraphValue in buildResolvedIncidentNode) and stop sanitising inside loadIncidentAnalysis.
- min_similarity_score: agreed it’s for the read/search path; I’ll add a short inline comment in the config gate noting it’s intentionally unused in the write path.
- Magic numbers: agreed—I'll replace the literals with named constants (e.g., maxGraphLogLines = 3, maxGraphCommits = 2) so the limits are explicit.
I’ll push a small follow-up commit with these tidy-ups.

ntcdeveloper (approved):
All three points addressed in the follow-up commit. Merging.

Issue #29, opened 2026-05-06
improvement: Grafana MCP server as optional team-level data source


ntcdeveloper:
## Idea

Grafana ships an MCP (Model Context Protocol) server that allows AI models to query Grafana data sources directly. Teams that already use Grafana have all their data sources configured there — Prometheus, Loki, CloudWatch, Tempo, etc. A single Grafana MCP connection would give Wachd access to all of them without requiring separate credentials per source.

## Why this is useful

Today a team admin must configure each data source individually in Wachd (`team_data_sources`): Prometheus endpoint + auth, Loki endpoint + auth, etc. If the team already has Grafana, those connections are already set up there. A Grafana MCP integration lets the team connect once and cover everything Grafana can already query.

This also pairs well with #27 (service dependency graph) — querying infra-level metrics for sibling services becomes a single MCP call rather than multiple separate API connections.

## Design

- New data source type: `grafana_mcp` in `team_data_sources`
- Credentials: Grafana instance URL + service account token with viewer scope
- Collector checks for a `grafana_mcp` data source first; if present, uses it for metric and log queries
- Falls back to direct API connections (Prometheus, Loki, etc.) if not configured
- Fully optional — teams without Grafana are unaffected

## Notes

- Grafana MCP server must be enabled on the team's Grafana instance (self-hosted or Grafana Cloud)
- Not a replacement for direct API connectors — Datadog-only and Splunk-only teams won't have Grafana
- Reduces setup friction significantly for teams already on Grafana (which is most teams)
- Related: #27 — dependency graph queries become easier with MCP as a unified query layer

## References

- Grafana MCP server: https://grafana.com/docs/grafana/latest/developers/mcp/

PR #51, opened 2026-05-23
feat(collector): add optional Grafana MCP datasource

ADITYA-CODE-SOURCE:
Addresses #29

This adds Grafana MCP as an optional team-level datasource so teams can query logs and metrics through a single Grafana connection instead of configuring Prometheus and Loki separately.

What changed
- Added `grafana_mcp_url` and encrypted `grafana_mcp_token` fields to team config storage and team config API responses.
- Added a Grafana MCP collector that initializes an MCP session over JSON-RPC and queries Loki/Prometheus tools through Grafana.
- Worker now prefers Grafana MCP for logs and metrics when configured, and falls back to direct Loki/Prometheus connectors if MCP is unavailable or errors.
- Added unit tests for the collector, worker MCP selection/fallback, and DB-backed team config round-trip.

Verification
- `GOFLAGS=-p=1 go test ./cmd/worker`
- `GOFLAGS=-p=1 go test ./internal/collector`
- `GOFLAGS=-p=1 go test ./internal/store`
- `GOFLAGS=-p=1 go test ./cmd/server`

Note
- I hit Windows memory pressure when running `go test ./...` in one shot in this environment, so I verified the changed packages individually with reduced parallelism.


ADITYA-CODE-SOURCE:
@ntcdeveloper  @fatkobra  
Quick note on design:
- I implemented Grafana MCP config in team_config (instead of adding a new team_data_sources table) because the current repo stores datasource settings there already (Prometheus/Loki/Dynatrace/Splunk/GitHub). This keeps the change incremental and consistent with existing patterns.
- If you’d prefer to follow the team_data_sources design from the issue strictly, I’m happy to refactor/split this into a follow-up once we agree on the direction.

Happy to adjust based on your preferred architecture.

ntcdeveloper (commented):
Good feature — SSRF validation on the MCP URL, token encryption following the same pattern as GitHub/Dynatrace/Splunk, and the fallback logic are all correct. One thing needs fixing before merge, two others for discussion.

**Required fix — VARCHAR(500) vs 2000-char server validation mismatch**
`schema.sql` defines `grafana_mcp_url VARCHAR(500)` but the server validates `len > 2000`. A URL between 501–2000 chars passes the server check then hits a DB truncation or error. The other URL columns (`prometheus_endpoint`, `loki_endpoint`) are also 500 — either align the server check down to 500 or bump the column to match.

---

**Discussion — `duration.String()` in PromQL expression**
`fmt.Sprintf(`rate(http_errors_total{service=%q}[%s])`, service, duration)` — `duration.String()` on `30 * time.Minute` produces `30m0s`. That's valid in recent Prometheus builds but fragile against older instances. Something like `fmt.Sprintf("%ds", int(duration.Seconds()))` is more portable.

**Discussion — `lookupDatasourceUID` called twice per incident**
The datasource list is fetched once to get the Loki UID and again to get the Prometheus UID. Both calls hit the same `list_datasources` tool with no caching in between. Easy to fetch once, build a map, then look up both UIDs from it — saves a round trip on every alert.

Neither discussion item is a blocker. The VARCHAR mismatch is.

ADITYA-CODE-SOURCE:
> Good feature — SSRF validation on the MCP URL, token encryption following the same pattern as GitHub/Dynatrace/Splunk, and the fallback logic are all correct. One thing needs fixing before merge, two others for discussion.
> 
> **Required fix — VARCHAR(500) vs 2000-char server validation mismatch** `schema.sql` defines `grafana_mcp_url VARCHAR(500)` but the server validates `len > 2000`. A URL between 501–2000 chars passes the server check then hits a DB truncation or error. The other URL columns (`prometheus_endpoint`, `loki_endpoint`) are also 500 — either align the server check down to 500 or bump the column to match.
> 
> **Discussion — `duration.String()` in PromQL expression** `fmt.Sprintf(`rate(http_errors_total{service=%q}[%s])`, service, duration)` — `duration.String()` on `30 * time.Minute` produces `30m0s`. That's valid in recent Prometheus builds but fragile against older instances. Something like `fmt.Sprintf("%ds", int(duration.Seconds()))` is more portable.
> 
> **Discussion — `lookupDatasourceUID` called twice per incident** The datasource list is fetched once to get the Loki UID and again to get the Prometheus UID. Both calls hit the same `list_datasources` tool with no caching in between. Easy to fetch once, build a map, then look up both UIDs from it — saves a round trip on every alert.
> 
> Neither discussion item is a blocker. The VARCHAR mismatch is.

Thanks for the review
- Agreed on the VARCHAR(500) vs 2000 server validation mismatch — I’ll align them (I’ll update the server validation to 500 to match the existing Prometheus/Loki URL columns, unless you’d prefer bumping the DB column instead).
- Re: PromQL duration formatting: good point. I’ll switch to a more portable format (e.g., 1800s for 30 minutes).
- Re: datasource UID lookup: agreed — I’ll fetch datasources once per incident, build a map, and reuse it for both Loki/Prometheus.

I’ll also rebase onto the latest main and push an updated commit.

ntcdeveloper:
Confirmed — align the server validation down to 500 to match `prometheus_endpoint` and `loki_endpoint`. No need to bump the DB column. All three fixes look good, waiting on the updated commit.

ADITYA-CODE-SOURCE:
Pushed follow-up a58fd8f:
- aligned grafana_mcp_url validation to 500 chars (matching VARCHAR(500) + Prometheus/Loki)
- formatted PromQL range duration as seconds for portability
- cached MCP datasource lookup so list_datasources is called once per session
Also rebased onto latest main.

PR #53, opened 2026-05-24
feat(collector): add service dependency context collection

### LINKED ISSUE #27: improvement: service dependency graph for cross-service root cause analysis
## Problem

Context collection is currently scoped to the alerting service — its commits, logs, and metrics. If the actual root cause is a shared dependency (Redis, message bus, Postgres, load balancer), the AI works from incomplete evidence and can produce a confident but wrong diagnosis.

This is one of the most common real-world failure patterns: service A alerts, but Redis or the message bus is the actual cause.

Raised in feedback from the r/Observability community.

## Proposed solution

Let users declare service relationships per team. When an alert fires for service A, the collector also pulls metrics and logs for its declared dependencies and includes them in the causal timeline.

### New table

```sql
CREATE TABLE service_dependencies (
    id          UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    team_id     UUID NOT NULL REFERENCES teams(id) ON DELETE CASCADE,
    service     TEXT NOT NULL,   -- matches the service tag on incoming alerts
    depends_on  UUID NOT NULL REFERENCES team_data_sources(id) ON DELETE CASCADE,
    label       TEXT,            -- human label, e.g. "shared Redis", "payments Postgres"
    created_at  TIMESTAMPTZ NOT NULL DEFAULT now()
);
CREATE INDEX ON service_dependencies(team_id, service);
```

### Collector change

When processing an alert for a service, after loading the team's primary data sources, also load `service_dependencies WHERE team_id = $1 AND service = $2` and pull metrics/logs from each declared dependency.

### Correlator change

Include the dependency timeline (metric spikes, error bursts) in the causal context sent to the AI alongside the alerting service's timeline.

### API

```
GET  /api/v1/teams/{teamId}/dependencies
POST /api/v1/teams/{teamId}/dependencies
DELETE /api/v1/teams/{teamId}/dependencies/{id}
```

Team admin self-service — no superadmin required.

## Notes

- Fits cleanly into the existing per-team data source model — no architectural rewrite
- Relationships are user-declared (not inferred from traces), which keeps it instrumentation-free
- Start with infra-type dependencies (Redis, Postgres, message bus) — service-to-service tracing is a separate problem
- Needs design review before implementation

ntcdeveloper:
## Summary

- Adds `service_dependencies` table (scoped by `team_id`, self-dep guard, `UNIQUE(team_id, service, depends_on)`)
- `store.ListServiceDependencies`, `CreateServiceDependency`, `DeleteServiceDependency` — all filter by `team_id`
- REST API under `/api/v1/teams/{teamId}/dependencies`:
  - `GET` — list all deps, optional `?service=` filter
  - `POST` — create a dependency (admin only)
  - `DELETE /{id}` — remove a dependency (admin only)
- Worker: `collectDependencyContext` loads deps from DB and calls `applyDependencyContext` (pure function)
- `applyDependencyContext` fetches Loki logs + Prometheus metrics for each dependency service, with SSRF validation per endpoint before any HTTP call
- `TODO(#29)`: when Grafana MCP (PR #51) merges, swap in `collectGrafanaMCPContext` inside the dep loop

## Test plan

- [ ] `go build ./...` — clean
- [ ] `go vet ./...` — clean
- [ ] `TestApplyDependencyContext_*` (6 unit tests in `cmd/worker/`) — all pass
- [ ] `TestDB_ServiceDependencies_*` (4 integration tests in `internal/store/`) — all pass
- [ ] SSRF test: Loki endpoint set to `169.254.169.254` — blocked, no logs collected
- [ ] Cross-team delete: team B cannot delete team A's dependency (RowsAffected check)

Closes #27

### DIFF
--- a/cmd/server/e2e_integration_test.go
+++ b/cmd/server/e2e_integration_test.go
@@ -794,3 +794,75 @@ func TestE2E_UpdateMember_UserMustBeTeamMember(t *testing.T) {
 		t.Errorf("phone update for non-member: got %d, want 422", resp.Code)
 	}
 }
+
+// TestE2E_ServiceDependencies_FilterByService verifies that GET
+// /api/v1/teams/{teamId}/dependencies?service=<name> returns only dependencies
+// where service matches, and that omitting the filter returns all dependencies.
+func TestE2E_ServiceDependencies_FilterByService(t *testing.T) {
+	env := newE2EEnv(t)
+
+	createDep := func(service, dependsOn string) {
+		t.Helper()
+		req := httptest.NewRequest("POST",
+			fmt.Sprintf("/api/v1/teams/%s/dependencies", env.kongTeam.ID),
+			e2eBody(t, map[string]any{"service": service, "depends_on": dependsOn}))
+		req.Header.Set("Content-Type", "application/json")
+		req.AddCookie(env.kongCookie)
+		resp := env.do(req)
+		if resp.Code != http.StatusCreated {
+			t.Fatalf("create dep %s→%s: got %d, want 201", service, dependsOn, resp.Code)
+		}
+	}
+
+	createDep("checkout-api", "redis-cache")
+	createDep("checkout-api", "payments-db")
+	createDep("payments-service", "postgres")
+
+	listDeps := func(serviceFilter string) []map[string]any {
+		t.Helper()
+		url := fmt.Sprintf("/api/v1/teams/%s/dependencies", env.kongTeam.ID)
+		if serviceFilter != "" {
+			url += "?service=" + serviceFilter
+		}
+		req := httptest.NewRequest("GET", url, nil)
+		req.AddCookie(env.kongCookie)
+		resp := env.do(req)
+		if resp.Code != http.StatusOK {
+			t.Fatalf("list deps (filter=%q): got %d, want 200", serviceFilter, resp.Code)
+		}
+		var deps []map[string]any
+		if err := json.NewDecoder(resp.Body).Decode(&deps); err != nil {
+			t.Fatalf("decode deps: %v", err)
+		}
+		return deps
+	}
+
+	// No filter — all three deps returned.
+	all := listDeps("")
+	if len(all) < 3 {
+		t.Errorf("expected at least 3 deps with no filter, got %d", len(all))
+	}
+
+	// Filter by checkout-api — only its two deps returned.
+	checkoutDeps := listDeps("checkout-api")
+	if len(checkoutDeps) != 2 {
+		t.Errorf("expected 2 deps for checkout-api, got %d", len(checkoutDeps))
+	}
+	for _, d := range checkoutDeps {
+		if d["service"] != "checkout-api" {
+			t.Errorf("filter returned dep with service=%q, want checkout-api", d["service"])
+		}
+	}
+
+	// Filter by payments-service — only its one dep returned.
+	paymentsDeps := listDeps("payments-service")
+	if len(paymentsDeps) != 1 {
+		t.Errorf("expected 1 dep for payments-service, got %d", len(paymentsDeps))
+	}
+
+	// Filter by a depends_on value — should return nothing (filter is on service, not depends_on).
+	redisDeps := listDeps("redis-cache")
+	if len(redisDeps) != 0 {
+		t.Errorf("filter by depends_on value should return 0 results, got %d", len(redisDeps))
+	}
+}
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -596,6 +596,9 @@ func main() {
 	apiRouter.HandleFunc("/{teamId}/config/test-notification", server.handleTestNotification).Methods("POST")
 	apiRouter.HandleFunc("/{teamId}/escalation", server.handleGetEscalationPolicy).Methods("GET")
 	apiRouter.HandleFunc("/{teamId}/escalation", server.handleUpsertEscalationPolicy).Methods("PUT")
+	apiRouter.HandleFunc("/{teamId}/dependencies", server.handleListServiceDependencies).Methods("GET")
+	apiRouter.HandleFunc("/{teamId}/dependencies", server.handleCreateServiceDependency).Methods("POST")
+	apiRouter.HandleFunc("/{teamId}/dependencies/{depId}", server.handleDeleteServiceDependency).Methods("DELETE")
 
 	// Wrap router with CORS middleware
 	handler := corsMiddleware(router)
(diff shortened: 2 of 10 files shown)


ntcdeveloper:
@fatkobra @ADITYA-CODE-SOURCE please review when you get a chance — this closes #27 (service dependency context collection). Builds on the graph work from PRs #39/#40 and adds a TODO(#29) for the Grafana MCP hook once PR #51 merges.

ntcdeveloper:
Rebased onto main after PR #51 merged. Resolved the conflict in `helper.go` — both `collectGrafanaMCPContext` (from #51) and `collectDependencyContext`/`applyDependencyContext` (this PR) are present. Removed the stale TODO(#29) and replaced it with an architectural note. All tests pass.

ADITYA-CODE-SOURCE:
> @fatkobra @ADITYA-CODE-SOURCE please review when you get a chance — this closes #27 (service dependency context collection). Builds on the graph work from PRs #39/#40 and adds a TODO(#29) for the Grafana MCP hook once PR #51 merges.

LGTM — this is a solid implementation for #27.
- Data model looks right: service_dependencies is team-scoped, prevents self-deps, and enforces UNIQUE(team_id, service, depends_on).
- Store/API paths appear properly tenant-scoped (team_id filters) and admin-only for POST/DELETE.
- Worker logic is clean: collectDependencyContext + pure applyDependencyContext, with SSRF validation before any HTTP calls.
- Tests are high-signal: unit tests for applyDependencyContext, integration tests for CRUD + cross-team delete, and SSRF block coverage.

Non-blocking suggestion: add an explicit per-incident cap (and/or per-dependency timeout) so dependency fan-out can’t grow unbounded if a team configures many dependencies.

fatkobra:
Thanks for putting this together and sorry for being late. Overall I think this is the right direction for #27.

What looks good:

- `service_dependencies` is team-scoped.
- Store methods filter by `team_id`.
- POST/DELETE are admin-only, while GET only requires team access.
- Worker integration is cleanly separated into `collectDependencyContext` / `applyDependencyContext`.
- SSRF validation happens before dependency Loki/Prometheus calls.
- The store/API/worker split feels consistent with the rest of the codebase.

I did find a few things I think should be fixed or clarified before merging.

### 1. `GET /dependencies?service=` looks advertised but not implemented

The PR summary says the GET endpoint supports an optional `?service=` filter.

But the handler currently calls:

```go
deps, err := s.cfg.ListServiceDependencies(r.Context(), teamID, "")
```

So the API always lists all dependencies for the team, even when `?service=checkout-ap`i is provided.

I think this should pass the query param through:
```
service := strings.TrimSpace(r.URL.Query().Get("service"))

deps, err := s.cfg.ListServiceDependencies(r.Context(), teamID, service)
```
This is small, but I think it should be fixed before merge because it is part of the advertised API behavior.

2. Dependency collection should probably support Grafana MCP too

After the rebase on #51, both Grafana MCP collection and dependency context collection are present.

For the primary service, `collectContext` can use Grafana MCP first and then fall back to direct Loki/Prometheus.

For dependencies, `applyDependencyContext` currently only does direct Loki/Prometheus calls.

That means a team configured with only Grafana MCP may get primary-service context, but no dependency context.

I think dependency collection should mirror the primary service behavior:

- Try Grafana MCP for `dep.DependsOn` when configured.
- Fall back to direct Loki/Prometheus only when MCP is unavailable or unsuccessful.
- Avoid double-collecting the same logs/metrics when MCP already succeeded.

This matters because #27 says dependency context should use the team's existing configured connectors, and Grafana MCP is now one of those connectors.

3. Add a database-level self-dependency guard

`CreateServiceDependency` correctly rejects self-dependencies in Go.

But the schema itself does not appear to prevent direct DB inserts like:
```
service = 'api'
depends_on = 'api'
```
Since the PR summary mentions a self-dep guard, I think the DB should enforce it too.

Maybe add something like:
```
CHECK (lower(btrim(service)) <> lower(btrim(depends_on)))
```
I would also consider guarding empty/whitespace values at the DB layer:
```
CHECK (btrim(service) <> '')
CHECK (btrim(depends_on) <> '')
```
The app-level validation is good, but the DB constraint makes the invariant impossible to bypass later.

4. API validation/error mapping could be friendlier

Right now `handleCreateServiceDependency` checks only raw empty strings before calling the store.

So values like `" "` will fail in the store and come back as a 500.

Also, duplicate dependencies or self-dependencies likely come back as 500, even though they are client/input problems.

I think the handler should trim inputs before validation and return:

`400 Bad Request` for missing/blank service or depends_on
`400 Bad Request` for self-dependency
`409 Conflict` for duplicate dependency
`500 Internal Server Error` only for real unexpected failures

This is not a huge issue, but it makes the API nicer and avoids turning user mistakes into server errors.

5. Consider a dependency fan-out cap / timeout

This can be non-blocking, but I agree with the earlier comment that dependency fan-out should be bounded.

Each dependency can trigger Loki and Prometheus calls. If a team configures many dependencies, one incident could create a lot of outbound calls and slow the worker.

A simple v1 guard would be enough:

cap the number of dependencies processed per incident
use a per-dependency timeout
maybe log when dependencies are skipped because of the cap

Something like `maxDependencyContextServices = 10` would be a reasonable start.

6. Tests I would like to see added

The current tests are useful, especially the SSRF/no-panic cases.

I think two more tests would make this safer:

API/list test proving `GET /dependencies?service=...` actually filters results.
Dependency collection test for a Grafana MCP-only config, or at least a clear TODO/follow-up issue if MCP dependency collection is intentionally deferred.

Overall: strong foundation, but I would fix the query-param bug and clarify/handle the Grafana MCP dependency path before merging.

ntcdeveloper:
@fatkobra @ADITYA-CODE-SOURCE — feedback addressed in the latest commit. Here's what changed:

**Issue 1 — `?service=` filter (blocking):** Handler now reads and trims `r.URL.Query().Get("service")` and passes it to `ListServiceDependencies`. Added an e2e test (`TestE2E_ServiceDependencies_FilterByService`) that exercises the filter, including a case where filtering by a `depends_on` value returns 0 results.

**Issue 2 — Grafana MCP for dependencies (blocking):** `collectDependencyContext` now tries MCP per dep first when the team has MCP configured, using the same SSRF/decrypt guard as the primary service path. Deps where MCP returns an error fall back to direct Loki/Prometheus. An empty-but-no-error MCP response still marks the dep as handled (no double-collect). Added 4 unit tests for `collectDepViaMCP` covering success, partial failure, both-error, and empty-success cases.

**Issue 3 — DB self-dep guard:** Three `CHECK` constraints added to `service_dependencies` via idempotent `DO/EXCEPTION` blocks: `lower(btrim(service)) <> lower(btrim(depends_on))`, `btrim(service) <> ''`, `btrim(depends_on) <> ''`.

**Issue 4 — API error mapping:** Handler trims inputs before validation (whitespace → 400), adds an explicit self-dep check at handler level (400), and detects pgconn SQLSTATE `23505` unique violations (409 Conflict). Only unexpected errors return 500.

**Issue 5 — fan-out cap:** `maxDependencyContextServices = 10` applied in `collectDependencyContext` with a log line when truncated.

cmd/server/e2e_integration_test.go
```diff
@@ -794,3 +794,75 @@ func TestE2E_UpdateMember_UserMustBeTeamMember(t *testing.T) {
...
+	if len(paymentsDeps) != 1 {
+		t.Errorf("expected 1 dep for payments-service, got %d", len(paymentsDeps))
+	}
+
+	// Filter by a depends_on value — should return nothing (filter is on service, not depends_on).
+	redisDeps := listDeps("redis-cache")
+	if len(redisDeps) != 0 {
+		t.Errorf("filter by depends_on value should return 0 results, got %d", len(redisDeps))
+	}
+}
```
fatkobra, line 868: Overall: strong foundation.
THis looks good. All the feedback was now addressed on the latest submit

fatkobra (approved):
Overall: strong foundation.
This looks good. All the feedback was now addressed on the latest submit

Issue #3, opened 2026-04-19
feat: Config-as-Code — wachd.yaml + wachd CLI


ntcdeveloper:
## Summary

Add Config-as-Code support so all team configuration can live in a Git-tracked `wachd.yaml` file instead of the database. Incidents, audit logs, and snooze state stay in PostgreSQL — only configuration is declarative.

## Motivation

- **DR without DB replication**: config survives database loss — `git clone` + `wachd sync` restores all team config instantly
- **GitOps-native**: every config change is a PR with a diff, a reviewer, and a `git revert` for rollback
- **Self-service with guardrails**: teams manage their own directory, CODEOWNERS enforces ownership, CI validates before apply
- **No manual re-entry after cluster rebuild**: new cluster + `wachd sync wachd.yaml` and it's ready

## What goes in YAML (config — declarative)

- Teams + members
- On-call schedules + rotations + time windows
- Escalation policies
- Data sources (Loki, GitHub, Prometheus, Dynatrace, Splunk)
- Notification channels (Slack, Teams, email)
- Alert routing rules
- Webhook integrations

## What stays in PostgreSQL (events — always)

- Incidents (fired, acked, snoozed, resolved)
- Acknowledgement history
- Call + SMS logs
- Escalation state
- Audit trail

### Repo layout (one per org)

```
wachd-config/
├── teams/
│   ├── platform/
│   │   ├── team.yaml
│   │   ├── schedule.yaml
│   │   └── datasources.yaml
│   └── payments/
│       └── team.yaml
└── .github/
    └── CODEOWNERS   # payments/ @payments-leads
```

### Secret references (never plaintext in Git)

```yaml
channels:
  slack:
    token_ref: k8s:wachd-secrets:slack-token
datasources:
  - type: loki
    credentials_ref: k8s:wachd-secrets:loki-token
```

### CLI

```bash
wachd validate wachd.yaml   # schema + cross-ref validation
wachd diff wachd.yaml       # show what would change
wachd sync wachd.yaml       # apply to running instance via API
wachd dump > wachd.yaml     # export current DB state → migrate to GitOps
```

### CI pipeline

```yaml
on: push to main
jobs:
  apply:
    - wachd validate wachd.yaml
    - wachd sync wachd.yaml --server https://wachd.company.internal --token $CI_TOKEN
```

## Foundation already in place

[...]

Issue #2, opened 2026-04-18
fix: security audit findings — SSRF gaps, SQL pattern, Next.js DoS


ntcdeveloper:
## Security Review — 2026-04-18

Findings from internal security audit. All items must be resolved before next release.

---

### HIGH — Next.js 16.2.2 DoS

- **File:** `web/package.json`
- **CVE:** [GHSA-q4gf-8mx6-v5v3](https://github.com/advisories/GHSA-q4gf-8mx6-v5v3) — Denial of Service via Server Components
- **Fix:** upgrade to Next.js 16.2.4

---

### MEDIUM — `fmt.Sprintf` table name in parameterized query

- **File:** `internal/store/team_members.go:117`
- Table name is built with `fmt.Sprintf` even though it comes from a `switch`. Safe today, but a dangerous pattern.
- **Fix:** replace with two explicit queries, one per table.

---

### MEDIUM — `validateEndpointURL` not called at collection time

- **Files:** `cmd/worker/helper.go:90, 102, 122, 166`
- SSRF validation runs on config write but collector endpoints are consumed from the DB at runtime with no re-validation. Config inserted before the validator existed can contain internal URLs.
- **Fix:** call `validateEndpointURL` in the worker before instantiating each collector.

---

### LOW — SSRF bypass edge cases in `validateEndpointURL`

- **File:** `cmd/server/main.go:419–453`
- Blocklist misses `0.0.0.0`, IPv4-mapped IPv6 (`::ffff:127.0.0.1`), `localhost.localdomain`
- **Fix:** add to blocked hostnames/ranges.

---

### Checklist

- [ ] Upgrade Next.js to 16.2.4
- [ ] Fix `UpdateMemberPhone` — remove `fmt.Sprintf` table interpolation
- [ ] Add `validateEndpointURL` in worker before each collector is instantiated
- [ ] Harden `validateEndpointURL` with missing bypass cases

Issue #1, opened 2026-04-18
arch: move AI backend config from team_config to system-level (superadmin)


ntcdeveloper:
## Problem

ai_backend and ai_model are currently stored in team_config (per-team), which is architecturally wrong.

In the enterprise model, the AI backend is a platform decision made by the superadmin — not a per-team choice. The enterprise picks Claude, OpenAI, or Ollama for the whole platform. Individual teams do not own the API key, do not manage the Ollama cluster, and should not be able to switch backends independently.

## Current (wrong) state

- AIBackend / AIModel fields on store.TeamConfig (internal/store/models.go:144)
- Stored in team_config table, one row per team
- Configurable via PUT /api/v1/teams/{teamId}/config (team-level endpoint)
- Worker loads AI backend from team_config per incident

## Correct design

Superadmin sets AI backend once for the platform via POST /api/v1/admin/system/ai.
All teams use the same backend. API key is read from K8s secret, never stored in DB.

## Required changes

1. Remove ai_backend, ai_model from team_config table; add singleton system_config table
2. Remove AIBackend/AIModel from store.TeamConfig; add SystemConfig struct
3. Add store/system_config.go with GetSystemConfig / UpsertSystemConfig
4. Remove AI fields from handleUpsertTeamConfig
5. Add superadmin routes: GET/PUT /api/v1/admin/system/ai
6. Worker loads AI backend from system_config, not team_config
7. values.yaml ai.backend seeds system_config on first startup only — DB is authoritative after

## Acceptance criteria

- PUT /api/v1/teams/{teamId}/config no longer accepts or returns ai_backend / ai_model
- GET/PUT /api/v1/admin/system/ai exists and requires superadmin role
- Worker reads AI backend from system_config, not team_config
- Existing team_config rows are migrated (drop columns)
- READMEs updated: superadmin sets AI backend once for the platform
- CLAUDE.md updated to reflect system-level AI config

ntcdeveloper:
Implemented in e619a9c — platform-wide system config is now a singleton table managed by superadmin only. New endpoints: GET/PUT /api/v1/admin/system/ai. Worker seeds from env vars on first startup, reads from DB on subsequent runs. All 20 e2e tests pass. Tagged v0.3.1.

Issue #42, opened 2026-05-22
feat(graph): add REST endpoints for incident similarity and subgraph


ntcdeveloper:
## Context

`internal/graph.PostgresStore` implements `FindSimilar` and `GetSubgraph` but neither is exposed via the API. No REST endpoint exists. This blocks the frontend panel and the notification integration.

## Endpoints to implement

```
GET  /api/v1/teams/{teamId}/incidents/{id}/similar
GET  /api/v1/teams/{teamId}/incidents/{id}/graph
POST /api/v1/teams/{teamId}/incidents/{id}/promote
DELETE /api/v1/teams/{teamId}/graph/nodes/{nodeId}
GET  /api/v1/teams/{teamId}/graph/config
PUT  /api/v1/teams/{teamId}/graph/config
```

**RBAC:**
- `GET /similar`, `GET /graph` — `viewer` and above
- `POST /promote`, `DELETE /nodes/{id}` — `admin` only
- `GET|PUT /graph/config` — `admin` only

All endpoints must validate `teamId` and enforce team isolation — the same pattern as every other team-scoped endpoint in `main.go`.

**Response shapes:**

`GET /similar`:
```json
{
  "data": [
    {
      "incident_id": "uuid",
      "title": "Payment timeout",
      "score": 0.84,
      "reason": "similar root cause; same service: checkout-api",
      "occurred_at": "2026-03-12T03:14:00Z",
      "resolution": "rolled back v2.3.1"
    }
  ]
}
```

`GET /graph/config`:
```json
{
  "enabled": true,
  "min_similarity_score": 0.12
}
```

## Schema

Add the team graph config table to `internal/store/schema.sql`:

```sql
CREATE TABLE IF NOT EXISTS team_graph_config (
    team_id              UUID PRIMARY KEY REFERENCES teams(id) ON DELETE CASCADE,
    enabled              BOOLEAN NOT NULL DEFAULT true,
    min_similarity_score FLOAT   NOT NULL DEFAULT 0.12,
    updated_at           TIMESTAMPTZ NOT NULL DEFAULT now()
);
```

## Testing requirements

Every new handler must have both unit and integration tests — this is a non-negotiable project rule.

**Unit tests:**
- `GET /similar` with no graph nodes returns empty array (not 404)
- `GET /similar` with wrong `teamId` returns 403
- `POST /promote` by a `viewer` returns 403
- `PUT /graph/config` with `min_similarity_score` outside 0.0–1.0 returns 400
[...]

PR #54, opened 2026-05-25
feat(graph): expose similarity and subgraph REST endpoints

ADITYA-CODE-SOURCE:
fix#42

This exposes the incident knowledge graph over the team API so similarity, subgraph traversal, manual promotion, node deletion, and graph configuration can be consumed by notifications and the frontend.

What changed
- Added graph endpoints for similar incidents, incident subgraphs, node promotion, node deletion, and graph config.
- Enforced viewer/admin RBAC and team isolation on every graph endpoint.
- Added graph config handling using the existing `team_graph_config` store defaults and validation.
- Added unit tests, HTTP integration tests, and graph API docs.

Verification
- `GOFLAGS=-p=1 go test ./cmd/server ./internal/graph`
- `GOFLAGS=-p=1 go test -tags integration -run TestE2E_Graph ./cmd/server`

### DIFF
--- a/cmd/server/e2e_integration_test.go
+++ b/cmd/server/e2e_integration_test.go
@@ -896,4 +896,18 @@ func TestE2E_GraphTeamIsolationAndViewerPromoteForbidden(t *testing.T) {
 	if resp.Code != http.StatusForbidden {
 		t.Fatalf("cross-team delete: got %d want 403", resp.Code)
 	}
+
+	req = httptest.NewRequest("GET", fmt.Sprintf("/api/v1/teams/%s/graph/config", env.kafkaTeam.ID), nil)
+	req.AddCookie(env.kongCookie)
+	resp = env.do(req)
+	if resp.Code != http.StatusForbidden {
+		t.Fatalf("cross-team graph config read: got %d want 403", resp.Code)
+	}
+
+	req = httptest.NewRequest("GET", fmt.Sprintf("/api/v1/teams/%s/graph/config", env.kongTeam.ID), nil)
+	req.AddCookie(env.kongViewerCookie)
+	resp = env.do(req)
+	if resp.Code != http.StatusOK {
+		t.Fatalf("viewer graph config read: got %d want 200 body=%s", resp.Code, resp.Body.String())
+	}
 }
--- a/cmd/server/e2e_integration_test.go
+++ b/cmd/server/e2e_integration_test.go
@@ -50,6 +50,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/redis/go-redis/v9"
 	"github.com/wachd/wachd/internal/auth"
+	"github.com/wachd/wachd/internal/graph"
 	"github.com/wachd/wachd/internal/license"
 	"github.com/wachd/wachd/internal/oncall"
 	"github.com/wachd/wachd/internal/queue"
@@ -145,6 +146,7 @@ func newE2EEnv(t *testing.T) *e2eEnv {
 		sessions:       sessions,
 		license:        lic,
 		enc:            enc,
+		graphStore:     graph.NewPostgresStore(db.Pool()),
 		webhookLimiter: newWebhookIPLimiter(),
 	}
 
@@ -172,6 +174,12 @@ func newE2EEnv(t *testing.T) *e2eEnv {
 	api.HandleFunc("/{teamId}/members", server.handleListMembers).Methods("GET")
 	api.HandleFunc("/{teamId}/config", server.handleGetTeamConfig).Methods("GET")
 	api.HandleFunc("/{teamId}/config", server.handleUpsertTeamConfig).Methods("PUT")
+	api.HandleFunc("/{teamId}/graph/config", server.handleGetGraphConfig).Methods("GET")
+	api.HandleFunc("/{teamId}/graph/config", server.handleUpsertGraphConfig).Methods("PUT")
+	api.HandleFunc("/{teamId}/graph/nodes/{nodeId}", server.handleDeleteGraphNode).Methods("DELETE")
+	api.HandleFunc("/{teamId}/incidents/{incidentId}/similar", server.handleGetSimilarIncidents).Methods("GET")
+	api.HandleFunc("/{teamId}/incidents/{incidentId}/graph", server.handleGetIncidentGraph).Methods("GET")
+	api.HandleFunc("/{teamId}/incidents/{incidentId}/promote", server.handlePromoteIncidentGraphNode).Methods("POST")
 	api.HandleFunc("/{teamId}/schedule/overrides", server.handleCreateOverride).Methods("POST")
 	api.HandleFunc("/{teamId}/members/{userId}", server.handleUpdateMember).Methods("PUT")
 
@@ -707,6 +715,7 @@ func TestE2E_Override_ViewerIsDenied(t *testing.T) {
 		t.Errorf("viewer create override: got %d, want 403", resp.Code)
 	}
 }
+
 // admin role — a responder must receive 403 on both GET and PUT.
 func TestE2E_Config_RequiresAdmin(t *testing.T) {
 	env := newE2EEnv(t)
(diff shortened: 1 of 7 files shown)


ntcdeveloper:
Thorough review done. Three required fixes before merge, two discussion points for @fatkobra to weigh in on.

---

**Required fix #1 — `lookupIncidentGraphNode` bypasses the graph Store interface**

This is the most important structural issue. The function runs raw SQL directly against `s.db.Pool()` inside `main.go`:

```go
err := s.db.Pool().QueryRow(ctx, `
    SELECT id, team_id, type, status, label, external_id, properties, created_at, updated_at
    FROM graph_nodes WHERE team_id = $1 AND type = 'incident' AND external_id = $2
`, teamID, incidentID.String()).Scan(...)
```

`postgres_store.go` already has `getNode` and `scanNode` with the correct scan order. By duplicating the logic here, if `graph_nodes` ever gets a new column or a reorder, this function returns silently wrong data — no compile error, no test failure. Fix: add `FindNodeByExternalID(ctx context.Context, teamID uuid.UUID, nodeType NodeType, externalID string) (*Node, error)` to the `graph.Store` interface, implement it in `PostgresStore`, and replace `lookupIncidentGraphNode` with a call through `s.graphStore`.

---

**Required fix #2 — `strings.Contains(err.Error(), "no rows")` should be `errors.Is(err, pgx.ErrNoRows)`**

Same function. `postgres_store.go` line 508 already shows the correct pattern — follow it:
```go
if errors.Is(err, pgx.ErrNoRows) {
    return nil, nil
}
```
String matching on error messages is fragile and not the pgx convention used everywhere else in this codebase.

---

**Required fix #3 — `handleGetGraphConfig` blocks viewers with `requireTeamAdmin`**

`GET /graph/config` is a read operation. Using `requireTeamAdmin` means a viewer or responder can't check whether the graph is enabled for their team. Should be `requireTeamAccess`, consistent with how other GET endpoints are guarded.

---

**Discussion for @fatkobra — N+1 queries in `handleGetSimilarIncidents`**

Each similarity match calls `GetIncident` individually — up to 10 round-trips per request. Acceptable at MVP scale but worth a `SELECT … WHERE id = ANY($1)` batch fetch before this goes to high traffic.

**Discussion for @fatkobra — `writeDataEnvelope` is inconsistent with the existing server style**

The graph endpoints now respond with `{"data": ..., "error": null}` while every other handler in `main.go` encodes the response directly. CLAUDE.md specifies the envelope as the standard, but the rest of the API doesn't follow it yet. Either the graph endpoints should match current style for consistency, or we align on migrating everything — good to get @fatkobra's call on this before it grows.

fatkobra:
Thanks for this PR. Overall I think this is the right direction for #42.

The endpoint set matches what we need for the frontend/notification path:

- similar incidents
- incident subgraph
- promote graph node
- delete graph node
- graph config read/write

I also like that the handlers are team-scoped and that the PR includes docs plus integration coverage.

I agree with the required fixes already called out by @ntcdeveloper, and I would treat these as blocking before merge.

### Required before merge

#### 1. Please move `lookupIncidentGraphNode` behind the graph store interface

I agree that `lookupIncidentGraphNode` should not run raw SQL directly from `main.go`.

The graph package already owns node scanning and graph persistence, so the server handler should not duplicate:

- SQL column order
- scan order
- graph node lookup rules
- pgx no-row handling

I would add something like this to `graph.Store`:

    FindNodeByExternalID(ctx context.Context, teamID uuid.UUID, nodeType NodeType, externalID string) (*Node, error)

Then implement it in `PostgresStore` using the existing `scanNode` helper.

After that, the handlers can lookup the incident node through `s.graphStore` instead of reaching into `s.db.Pool()`.

This keeps `main.go` as API orchestration code and keeps graph storage details inside `internal/graph`.

#### 2. Please avoid `strings.Contains(err.Error(), "no rows")`

This should use pgx/error wrapping instead of string matching.

Either the new store method can return `graph.ErrNodeNotFound`, or the Postgres implementation can use:

    errors.Is(err, pgx.ErrNoRows)

Then the handler can treat not-found as an empty graph/similar response or a 404 depending on the endpoint.

String matching on `"no rows"` is fragile and inconsistent with the rest of the graph store.

#### 3. I agree that `GET /graph/config` should be readable by team members

Even though the original issue listed graph config under admin-only, I think `GET /graph/config` should use `requireTeamAccess`.

The returned fields are not sensitive:

- `enabled`
- `min_similarity_score`

A viewer/responder may need to know whether the graph panel is enabled or why similarity results are unavailable.

My preference:

- `GET /graph/config` → viewer and above
- `PUT /graph/config` → admin only

If we make that change, please also update the docs and tests so they match.

### Discussion points

#### 4. N+1 incident lookups in `handleGetSimilarIncidents`

I agree this is not a huge issue for the current hardcoded limit of 10, but it is still worth fixing if this PR is already being updated.

Right now each similarity match does one `GetIncident` call. A batch fetch would be cleaner:

- collect all candidate incident IDs from `match.Node.ExternalID`
- fetch incidents in one query with `WHERE id = ANY($1)` and `team_id = $2`
- map results by ID
- build the response from that map

This would also make the handler easier to scale if the frontend later asks for more than 10 matches.

I would be okay with this as a follow-up if we keep the limit fixed at 10, but I would prefer fixing it now if the change is small.

#### 5. Response envelope

My vote is to keep the envelope for these graph endpoints.

Issue #42 explicitly asked for the standard success shape:

    {"data": ..., "error": null}

So I think this PR should follow that for the new graph API.

I would not try to migrate every existing endpoint in this PR. That should be a separate API consistency cleanup later.

Small docs wording suggestion: instead of saying all API responses use the envelope, say all graph endpoint success responses use the envelope. That avoids implying the whole existing API has already been migrated.

### Additional small things I noticed

#### 6. `graphStoreForTeam` silently ignores config load errors

`graphStoreForTeam` currently falls back to the default graph store if `GetTeamGraphConfig` returns an error.

That could hide a real DB/config problem and silently ignore a team's configured `min_similarity_score`.

I think it would be safer for this helper to return an error, for example:

    graphStoreForTeam(ctx, teamID) (graph.Store, error)

Then the handlers can return a 500 if graph config cannot be loaded.

#### 7. Clarify what `enabled` means

The schema comment says `team_graph_config.enabled` controls whether resolved incidents are written to the knowledge graph.

This PR exposes `enabled` in the REST API, but the read endpoints still return graph/similarity results regardless of that flag.

That may be fine, but I think it should be documented clearly.

Two possible meanings:

1. `enabled=false` stops future graph write-back only, but existing graph data remains readable.
2. `enabled=false` disables graph API results too.

I slightly prefer option 1, but the docs should say that so users are not surprised.

#### 8. Tests: avoid silently skipping handler coverage if possible

`graph_handlers_test.go` opens a real DB and skips if it cannot connect.

That is okay for integration-style coverage, but issue #42 asked for both unit and integration tests.

If possible, I would either:

- make these pure handler unit tests with a fake/mock graph store, or
- move DB-backed tests under the integration pattern and add at least one non-DB unit test for handler validation/RBAC behavior.

Not blocking if the maintainers are okay with the current test style, but silent skips can make coverage look stronger than it really is in CI.

### Summary

I am in favor of this PR after the required fixes.

The main thing I would not merge without is moving incident graph-node lookup into the graph store interface. That keeps the API layer clean and avoids duplicated SQL/scanning logic.

ADITYA-CODE-SOURCE:
Pushed follow-up 8f25137:
- moved incident node lookup behind graph.Store via FindNodeByExternalID
- removed the raw SQL / "no rows" string-matching path
- changed GET /graph/config to viewer+ access
- tightened graph config error handling
- batched incident fetches in /similar
- updated docs wording around the graph endpoint envelope and enabled semantics
- refreshed the graph tests

ntcdeveloper (approved):
Follow-up commit 8f25137 addresses all required fixes:

- `FindNodeByExternalID` on `graph.Store` interface — verified in `internal/graph/store.go` and `postgres_store.go`
- `errors.Is(err, pgx.ErrNoRows)` at lines 430 and 549 — no more string matching
- `GET /graph/config` uses `requireTeamAccess` — confirmed
- `graphStoreForTeam` returns `(graph.Store, error)` — confirmed
- Batch incident fetch uses `WHERE id = ANY($2)` — confirmed
- Docs and enabled semantics updated

@fatkobra all your required fixes are in — please take a final look when you get a chance.

fatkobra:
Thanks for the follow-up commit. I re-reviewed the updated PR and this addresses the main items I had concerns about.

What looks fixed now:

- `lookupIncidentGraphNode` was removed from `main.go`
- incident graph node lookup now goes through `graph.Store` via `FindNodeByExternalID`
- `PostgresStore` implements `FindNodeByExternalID` using the existing graph scan path
- not-found handling now uses `graph.ErrNodeNotFound` instead of string matching `"no rows"`
- `GET /graph/config` now uses `requireTeamAccess`
- `PUT /graph/config` remains admin-only
- the N+1 incident lookup in `handleGetSimilarIncidents` was replaced with a batch `getIncidentsByIDs` helper
- graph endpoint response envelopes were kept, which matches #42
- docs now clarify that only graph endpoint success responses use the envelope
- docs also clarify that `enabled=false` disables future graph write-back, while existing graph data remains readable
- extra E2E coverage was added for cross-team graph config read and viewer graph config read

I am good with this direction and I do not have any blocking comments from my side.

One tiny optional test nit: the current viewer promote forbidden E2E path appears to use a Kong viewer cookie against the Kafka team, so it proves cross-team denial. It would be slightly stronger to also test a viewer on their own team calling:

`POST /api/v1/teams/{teamId}/incidents/{incidentId}/promote`

and receiving `403`.

Not a blocker from my side because the handler uses `requireTeamAdmin`, but adding that would match the exact #42 test requirement more directly.

Overall: looks good to me after the follow-up commit.

PR #55, opened 2026-05-27
feat(worker): include similar incidents in notifications

### LINKED ISSUE #43: feat(graph): include similar past incident in alert notifications
## Context

When an alert fires, the notification (Slack, SMS, email) already includes probable cause and suggested action. It should also surface the most similar past incident when the similarity score is high enough — this is the highest-value touch point because the engineer sees it before opening the dashboard.

Depends on: issue for REST endpoints / graph API wiring.

## What to implement

In the alert dispatch path (after AI analysis, before `notify/`):

1. Call `graph.FindSimilar` for the new incident node (limit 1)
2. If the top result has score ≥ 0.50, append to the notification payload:
   - Similar incident title
   - Similarity score
   - Previous resolution (from node properties)
   - Link to the past incident in the dashboard
3. If `FindSimilar` returns an error or no results above threshold — continue without it. A plain alert is better than no alert (existing fail-safe rule).
4. Respect the team's `graph.enabled` config — skip if disabled.

**Slack message addition:**
```
🔁 Similar past incident (84%): Payment timeout · March 12
   Resolution: rolled back v2.3.1  →  [view]
```

**SMS:** append one short sentence only — SMS has a 160-char budget.

## Testing requirements

Every new function must have both unit and integration tests — this is a non-negotiable project rule.

**Unit tests:**
- Similar incident block is included when score ≥ 0.50
- Similar incident block is omitted when score < 0.50
- Similar incident block is omitted when `FindSimilar` returns an error (no crash, no alert drop)
- Similar incident block is omitted when team graph config has `enabled = false`
- SMS format stays within 160 characters when similar incident is appended

**Integration tests:**
- End-to-end: two similar incidents processed by the worker → second notification includes the first as a similar match
- Notification for a team with no graph history contains no similar incident block

## Acceptance criteria

- [ ] Slack and email notifications include similar incident when score ≥ 0.50
- [ ] SMS notification includes a short similar incident line within character budget
- [ ] `FindSimilar` failure never drops an alert
- [ ] Team graph config `enabled = false` suppresses the block
- [ ] All unit and integration tests pass: `go test ./...`

fatkobra:
## Summary

Closes #43.

This PR adds similar past incident context to alert notifications.

When an alert is processed, the worker now:

- builds a pending incident graph node after AI analysis
- respects `team_graph_config.enabled`
- calls `FindSimilar(..., limit=1)`
- includes the top match only when `score >= 0.50`
- safely skips similar incident context if graph lookup fails
- passes the optional similar incident into Slack, email, and SMS notifications

Notification behavior:

- Slack includes title, score, previous resolution, date, and dashboard link
- Email includes title, score, previous resolution, date, and dashboard link
- SMS appends one short similar-incident sentence and stays within 160 characters

The dashboard link uses `WACHD_DASHBOARD_URL` when set, and falls back to `http://localhost:3000`.

## Why

This gives the on-call engineer immediate historical context directly in the alert notification:

> This looks like a previous incident, and here is how it was resolved.

That means responders can see the likely previous fix before opening the dashboard.

## Fail-safe behavior

Similar incident lookup is best-effort only.

If graph config is disabled, no graph node is written and no similar block is added.

If graph lookup fails, notifications still send normally.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation

## Testing

Added unit coverage for:

- similar incident notification included when score is `>= 0.50`
- similar incident notification omitted when score is below `0.50`
- similar incident notification omitted when `FindSimilar` errors
- similar incident notification omitted when graph is disabled
- Slack similar incident block formatting
- email similar incident block formatting
- SMS similar incident formatting capped at 160 characters

Added worker integration coverage for:

- a second similar incident finding the first resolved incident through the graph
- a team with no graph history returning no similar incident context

Commands run:

```bash
go test ./internal/notify
go test ./cmd/worker
go test ./...
make test
go build ./...
go vet ./...
```

## Code Review Checklist

## Related Issues

Closes #43.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -78,3 +78,6 @@ LOKI_ENDPOINT=http://loki:3100
 # ── Logging ───────────────────────────────────────────────────────────────────
 LOG_LEVEL=debug
 ENVIRONMENT=development
+
+# Public web dashboard URL used in alert notification links.
+WACHD_DASHBOARD_URL=http://localhost:3000
--- a/cmd/worker/helper.go
+++ b/cmd/worker/helper.go
@@ -747,7 +747,7 @@ func (w *Worker) checkEscalationForIncident(ctx context.Context, incident *store
 	log.Printf("⬆ Escalating incident %s to step %d%s (%s): %s", incident.ID, nextStep, repeatLabel, nextLayer.LayerName, user.Name)
 	// Escalation re-notifications fire all immediate channels directly — user
 	// delay rules apply only to the initial contact, not escalation reminders.
-	w.fireAllChannels(ctx, incident, user, nil)
+	w.fireAllChannels(ctx, incident, user, nil, nil)
 }
 
 // resolveLayerMember returns the on-call member for a specific schedule at the
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -391,14 +391,16 @@ func (w *Worker) processAlertJob(ctx context.Context, job *queue.Job) error {
 		}
 	}
 
+	similarIncident := w.findSimilarIncidentForNotification(ctx, incident, sanitizedCtx, aiAnalysis)
+
 	// Notify on-call engineer via their personal notification rules (email/SMS/voice).
 	// Slack fires separately below — once per incident, team-level, not per-user.
 	if onCallUser != nil {
-		w.sendNotifications(ctx, incident, onCallUser, aiAnalysis)
+		w.sendNotifications(ctx, incident, onCallUser, aiAnalysis, similarIncident)
 	}
 
 	// Fire team Slack channel once — one post per incident, keeps the channel clean.
-	w.fireTeamSlack(ctx, incident, onCallUser, aiAnalysis)
+	w.fireTeamSlack(ctx, incident, onCallUser, aiAnalysis, similarIncident)
 
 	log.Printf("✓ Incident %s processed", incident.ID)
 	return nil
@@ -425,7 +427,7 @@ func (w *Worker) processResolvedIncidentJob(ctx context.Context, job *queue.Job)
 // If the user has notification rules configured, only those channels fire (with
 // optional delay). If no rules exist the function falls back to firing all
 // globally-configured channels immediately (preserving pre-rules behaviour).
-func (w *Worker) sendNotifications(ctx context.Context, incident *store.Incident, onCallUser *store.TeamMember, analysis *ai.AnalysisResponse) {
+func (w *Worker) sendNotifications(ctx context.Context, incident *store.Incident, onCallUser *store.TeamMember, analysis *ai.AnalysisResponse, similar *notify.SimilarIncident) {
 	rules, err := w.db.GetUserNotificationRules(ctx, onCallUser.ID, onCallUser.Source, "new_alert")
 	if err != nil {
 		log.Printf("warn: load notification rules for user %s: %v — falling back to all channels", onCallUser.ID, err)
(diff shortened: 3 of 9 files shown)


fatkobra:
@ntcdeveloper  @ADITYA-CODE-SOURCE 
Please review this one and provide feedback.

ntcdeveloper:
@fatkobra one issue before merge: `SlackNotifier.SendIncidentAlertWithSimilar` duplicates the full block-building logic from `SendIncidentAlert`. Email and SMS both do `if similar == nil { return e.SendIncidentAlert(...) }` and delegate cleanly — Slack should do the same. If `SendIncidentAlert` for Slack is updated later, the `WithSimilar` variant won't pick it up.

Also worth confirming: if `GetTeamGraphConfig` returns `nil` (no config row exists), the code falls through and uses the graph by default. Is the intended default "enabled" for teams with no config row?

fatkobra:
Very good catch!

Thanks, agreed on both points.

1. For Slack, I agree the current `SendIncidentAlertWithSimilar` duplicates too much of `SendIncidentAlert`.

I will refactor that so Slack follows the same pattern as email/SMS:

- if `similar == nil`, delegate directly to `SendIncidentAlert`
- extract the shared Slack alert message construction into a private helper
- have both `SendIncidentAlert` and `SendIncidentAlertWithSimilar` use that helper
- only add the similar-incident block in the `WithSimilar` path

That way future changes to the base Slack alert format are picked up by both paths.

2. For `GetTeamGraphConfig == nil`: yes, the intended default is enabled.

My reasoning is:

- `team_graph_config.enabled` defaults to `true` when a config row exists
- a missing row means the team has not explicitly opted out
- issue #43 says to skip when graph config has `enabled = false`

So the behavior should be:

- missing config row: enabled by default
- config row with `enabled = true`: enabled
- config row with `enabled = false`: disabled

I will make that intent explicit in the code/comment and add or update a test so the default-enabled behavior is documented instead of being hidden in `cfg != nil && !cfg.Enabled`.

fatkobra:
Follow-up refactor for the Slack path pushed.

What changed:

- `SlackNotifier.SendIncidentAlert` now builds the base Slack payload through a shared private helper.
- `SlackNotifier.SendIncidentAlertWithSimilar` now delegates directly to `SendIncidentAlert` when `similar == nil`, matching the email/SMS pattern.
- When a similar incident exists, the WithSimilar path reuses the same base Slack payload and only inserts the similar-incident block.
- Added a regression test proving the nil-similar Slack path produces the same payload as the base `SendIncidentAlert` path.

This should prevent the two Slack paths from drifting if the base Slack alert format changes later.

ntcdeveloper (approved):
Slack refactor looks good — `SendIncidentAlertWithSimilar` now delegates to `SendIncidentAlert` when `similar == nil`, shared helper is clean, and `insertSlackBlockBeforeIncidentID` is a nice touch for keeping the block order consistent.

One small follow-up still outstanding from your comment: the test documenting nil config = enabled-by-default. Not a blocker for merge but worth adding before the next release so the intent is captured.

fatkobra:
Thank you so much for the feedback, really appreciated.

I pushed a small follow-up commit documenting the nil-config behavior.

What changed:

- added a comment near the graph config check explaining that a missing config row means the team has not opted out, so graph lookup is enabled by default
- added `TestLoadSimilarIncidentForNotificationTreatsMissingGraphConfigAsEnabled`
- the test verifies that when `GetTeamGraphConfig` returns `nil`, the worker still builds the pending graph node, writes it, and calls `FindSimilar`
- the existing disabled-config test still verifies that an explicit `enabled=false` skips graph write and lookup

So the intended behavior is now covered:

- missing config row: enabled by default
- config row with `enabled=true`: enabled
- config row with `enabled=false`: disabled

PR #57, opened 2026-06-02
feat(web): add incident timeline view (#38)

### LINKED ISSUE #38: feature: incident timeline view
## Problem

During an incident, it is hard to understand what happened first, what changed, and what evidence points to the root cause.

## Proposed solution

Add an incident timeline view that shows key events in order.

The timeline could include:
- alert fired
- metric anomaly detected
- log pattern detected
- recent commits/deploys
- AI root cause analysis
- similar past incidents found
- escalation/notification events
- resolution

## Why this matters

A timeline makes the incident easier to understand for the on-call engineer and also makes postmortems easier to generate later.

This fits with issue #35 because graph edges like `triggered_after`, `preceded_by`, `resolved_by`, and `similar_to` can power the timeline.

### Alternatives Considered

Several alternatives were considered:

1. Show raw logs and alert events only  
This keeps all the information available, but it can be hard to understand during an incident. Engineers would still need to manually figure out what happened first and which events matter.

2. Rely only on the final AI summary  
A summary is helpful, but it hides the sequence of events. During debugging, teams often need to see the order of alerts, deploys, log patterns, AI analysis, and resolution steps.

3. Put timeline information only in the postmortem  
This helps after the incident, but not during the incident. A live timeline would help the on-call engineer understand what is happening while they are still debugging.

4. Use external monitoring dashboards only  
Existing dashboards are useful, but they usually show metrics or logs separately. They do not always combine alerts, deploys, commits, AI analysis, similar incidents, and resolution events into one incident-focused view.

5. Do nothing and let users piece events together manually  
This is the current workflow in many teams, but it slows down debugging and makes postmortems harder to write later.

### Feature Area

Incident management / timeline

ntcdeveloper:
## Summary

- New `GET /api/v1/teams/{teamId}/incidents/{incidentId}/timeline` endpoint that composes chronological events from incident fields, context JSONB (commits + first log spike), analysis JSONB, and sent notifications
- `GetIncidentNotificationEvents` store method — queries `pending_notifications` joined with users for display names, scoped by `team_id` + `incident_id`
- `IncidentTimeline` React component — vertical dot/line timeline with per-kind colours, loading skeleton, and meta badges
- `TimelineEvent` type added to `types.ts`; `api.incidents.timeline()` added to `api.ts`
- Timeline rendered above `SimilarIncidentsPanel` on the incident detail page

Closes #38

## Test plan

- [ ] `GET /api/v1/teams/{teamId}/incidents/{incidentId}/timeline` returns ordered events with correct `kind` values (`alert_fired`, `commit`, `log_spike`, `ai_analysis`, `notification_sent`)
- [ ] Timeline endpoint returns 404 for unknown incident ID
- [ ] Timeline endpoint returns 403 for incident belonging to a different team (tenant isolation)
- [ ] Frontend loading skeleton renders while request is in-flight
- [ ] Each event kind renders with correct colour and icon
- [ ] Component renders correctly when notification events are empty (no pending_notifications rows)
- [ ] `go build ./...` passes
- [ ] `go vet ./...` passes
- [ ] `cd web && npm run build` passes

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -30,6 +30,7 @@ import (
 	"net/netip"
 	"os"
 	"os/signal"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -607,6 +608,7 @@ func main() {
 	apiRouter.HandleFunc("/{teamId}/incidents/{incidentId}/similar", server.handleGetSimilarIncidents).Methods("GET")
 	apiRouter.HandleFunc("/{teamId}/incidents/{incidentId}/graph", server.handleGetIncidentGraph).Methods("GET")
 	apiRouter.HandleFunc("/{teamId}/incidents/{incidentId}/promote", server.handlePromoteIncidentGraphNode).Methods("POST")
+	apiRouter.HandleFunc("/{teamId}/incidents/{incidentId}/timeline", server.handleGetIncidentTimeline).Methods("GET")
 	apiRouter.HandleFunc("/{teamId}/dependencies", server.handleListServiceDependencies).Methods("GET")
 	apiRouter.HandleFunc("/{teamId}/dependencies", server.handleCreateServiceDependency).Methods("POST")
 	apiRouter.HandleFunc("/{teamId}/dependencies/{depId}", server.handleDeleteServiceDependency).Methods("DELETE")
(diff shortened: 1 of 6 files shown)


fatkobra:
This is very useful, great work here.

I request that: `error` should be reset when `teamId` / `incidentId` changes.

Right now, if one timeline request fails, `error` stays set. A later successful request updates `events`, but the component still renders the error branch because `error` is never cleared.

Could we reset state at the start of the effect, e.g.:

```ts
useEffect(() => {
  setLoading(true);
  setError(null);

  api.incidents.timeline(teamId, incidentId)
    .then(setEvents)
    .catch(() => setError('Failed to load timeline'))
    .finally(() => setLoading(false));
}, [teamId, incidentId]);
````



## Non-blocking comments


Non-blocking: should this show every sent notification row, or should it collapse retries / duplicates by user+channel?

Returning every `pending_notifications` row is useful for escalation history, but if retries create multiple sent rows for the same person/channel, the incident timeline may get noisy. So something to think about here!!!


Could you add a small test for timeline construction?

This endpoint has enough branching that a regression would be easy to miss:
- fired event is always present
- commits with timestamps are included
- only the first error/warn/critical log becomes `log_spike`
- analysis/ack/resolved/notifications are included when present
- response is sorted chronologically

A handler test with seeded incident context would give good coverage.

Optional: should the timeline include metric anomalies too?

The incident context/product model includes commits, logs, and metrics, but this first timeline pass only emits commit and log events from `incident.Context`. That’s fine as an MVP, but maybe worth either adding a metric event or tracking it as a follow-up.

fatkobra (changes requested):
change request only for the stale `error` state bug. The other stuff are non blockers.

ntcdeveloper:
Fixed in f6f72e5 — `setLoading(true)` and `setError(null)` now at the top of the effect before the fetch.

On the non-blockers:

**Notification deduplication** — good point, leaving it as-is for now. Returning every row is actually useful for escalation history (you can see retry #1, retry #2, who acked it). If it gets noisy in practice we can collapse by user+channel in a follow-up.

**Handler test** — agree it's worth having, will add in a follow-up PR rather than block this one further.

**Metric events** — tracked as a follow-up, nothing in the context JSONB maps cleanly to a point-in-time event yet.

fatkobra:
Great, no more blockers. Go ahead.

fatkobra (approved):
Looks good to me.

PR #58, opened 2026-06-02
feat(ai): make OpenAI base URL configurable

ntcdeveloper:
## Summary

The OpenAI backend had `api.openai.com` hardcoded, blocking any customer running an OpenAI-compatible inference server on their own infrastructure.

- `baseURL` field on `OpenAIBackend` — defaults to `https://api.openai.com/v1` when empty
- `OPENAI_BASE_URL` env var read in worker and passed through
- `analysis.openaiBaseURL` added to `values.yaml` (empty by default)
- `configmap.yaml` emits `OPENAI_BASE_URL` only when the value is set

With this change, operators can point Wachd's AI analysis at any OpenAI-compatible endpoint by setting one field:

```yaml
analysis:
  backend: openai
  openaiBaseURL: "http://llmkube-inference.default.svc.cluster.local/v1"
```

This works for: vLLM, LLMKube InferenceService, Azure OpenAI, Groq, Together.ai, Fireworks, or any self-hosted llama.cpp server.

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./internal/ai/...` passes
- [ ] Default behaviour unchanged: empty `openaiBaseURL` falls back to `api.openai.com/v1`
- [ ] Setting `analysis.openaiBaseURL` in values.yaml renders `OPENAI_BASE_URL` in the ConfigMap
- [ ] `OPENAI_BASE_URL` is absent from the ConfigMap when `openaiBaseURL` is empty

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -195,11 +195,12 @@ func main() {
 
 	case "openai":
 		apiKey := os.Getenv("OPENAI_API_KEY")
+		baseURL := os.Getenv("OPENAI_BASE_URL")
 		model := ""
 		if sc.AIModel != nil {
 			model = *sc.AIModel
 		}
-		aiEngine = ai.NewOpenAIBackend(apiKey, model)
+		aiEngine = ai.NewOpenAIBackend(apiKey, model, baseURL)
 		if aiEngine.IsAvailable(context.Background()) {
 			log.Printf("✓ OpenAI backend configured (model: %s)", aiEngine.GetModelName())
 		} else {
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -196,11 +196,12 @@ func main() {
 	case "openai":
 		apiKey := os.Getenv("OPENAI_API_KEY")
 		baseURL := os.Getenv("OPENAI_BASE_URL")
+		apiVersion := os.Getenv("OPENAI_API_VERSION")
 		model := ""
 		if sc.AIModel != nil {
 			model = *sc.AIModel
 		}
-		aiEngine = ai.NewOpenAIBackend(apiKey, model, baseURL)
+		aiEngine = ai.NewOpenAIBackend(apiKey, model, baseURL, apiVersion)
 		if aiEngine.IsAvailable(context.Background()) {
 			log.Printf("✓ OpenAI backend configured (model: %s)", aiEngine.GetModelName())
 		} else {
--- a/helm/wachd/README.md
+++ b/helm/wachd/README.md
@@ -66,10 +66,35 @@ Set `analysis.backend` to one of:
 |---|---|
 | `ollama` | Local LLM — air-gapped, no external API calls (default) |
 | `claude` | Anthropic Claude — highest quality analysis |
-| `openai` | OpenAI GPT-4o |
+| `openai` | OpenAI GPT-4o, or any OpenAI-compatible endpoint |
 | `gemini` | Google Gemini — free tier available |
 
 For air-gapped deployments, set `analysis.ollama.enabled: true` to deploy Ollama in-cluster alongside Wachd.
+
+### OpenAI-compatible endpoints
+
+`analysis.backend: openai` works with any OpenAI-compatible server — not just OpenAI itself. Set `analysis.openaiBaseURL` to point at a self-hosted inference server:
+
+```yaml
+# vLLM or LLMKube InferenceService (llama.cpp on 8080, vLLM on 8000)
+analysis:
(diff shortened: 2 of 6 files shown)


Defilan:
Took a pass through this from the LLMKube side. The change is clean: empty-string default to `api.openai.com/v1` keeps it backward compatible, the env var threads through the worker correctly, and gating `OPENAI_BASE_URL` in the ConfigMap on a non-empty value is the right call. Nothing needed on our end for this to work.

One correction for the values example so it works out of the box against LLMKube: an InferenceService's Service is a ClusterIP named after the InferenceService itself (not a fixed `llmkube-inference`), and the llama.cpp backend listens on 8080, not 80:

```yaml
analysis:
  backend: openai
  openaiBaseURL: "http://<inferenceservice-name>.<namespace>.svc.cluster.local:8080/v1"
```

(vLLM-backed services listen on 8000.) The exact URL is published on the resource at `.status.endpoint`, minus the trailing `/chat/completions`.

Minor optional nit: `Analyze` builds `o.baseURL + "/chat/completions"`, so a `baseURL` with a trailing slash yields a double slash. A `strings.TrimRight(baseURL, "/")` in the constructor would harden it. Not a blocker.

ntcdeveloper:
Both addressed in b2c34b6:

- Values example corrected to the accurate LLMKube format (llama.cpp on 8080, vLLM on 8000, with a note that the exact URL is at `.status.endpoint`)
- `strings.TrimRight(baseURL, "/")` added in the constructor

Good catch on both.

fatkobra:
The generic OpenAI-compatible path looks good, but I don’t think we should advertise Azure OpenAI support yet.

`Analyze` always sends `Authorization: Bearer <apiKey>`, while Azure OpenAI API-key auth expects the key in the `api-key` header; Bearer auth is for Entra tokens. That means the `Azure OpenAI` example in `values.yaml` / PR summary may not work for the common Azure API-key setup.

Can we either:
1. remove Azure OpenAI from the examples for this PR, or
2. add explicit Azure support, e.g. configurable auth header / provider mode?


Could we add a focused unit test for the new baseURL behavior?

The existing tests were updated to pass the third constructor arg, but they don’t assert:
- empty baseURL defaults to `https://api.openai.com/v1`
- custom baseURL is used
- trailing slash is trimmed so we don’t call `//chat/completions`

An `httptest.Server` around `Analyze` would cover this nicely.

**Optional docs:** `values.yaml` now documents `analysis.openaiBaseURL`, but the Helm README’s AI backend and key configuration sections still only describe `openai` as “OpenAI GPT-4o” and don’t mention the new `openaiBaseURL` option. ([[GitHub](https://github.com/wachd/wachd/pull/58/files)][1]) Since this is an operator-facing setting, I’d update the README too.

fatkobra (commented):
Can we either:

remove Azure OpenAI from the examples for this PR, or
add explicit Azure support, e.g. configurable auth header / provider mode?

ntcdeveloper:
Added in a9742df:

- `apiVersion` field — if non-empty, appended as `?api-version=...` (required for Azure OpenAI)
- Azure auth header auto-detection — URLs containing `.openai.azure.com` send `api-key: <key>` instead of `Authorization: Bearer`
- `OPENAI_API_VERSION` env var wired through worker + emitted from ConfigMap when set
- values.yaml now has a complete Azure example:

```yaml
analysis:
  backend: openai
  openaiBaseURL: "https://<resource>.openai.azure.com/openai/deployments/<deployment>"
  openaiAPIVersion: "2024-02-01"
```

Two new tests cover the auth header detection and the query param. Let me know if you'd prefer the Azure detection to be explicit (a provider field) rather than URL-based.

fatkobra:
This is looking good no blockers from me.

Good job.

ntcdeveloper:
Added in 653c8f6:

**Three baseURL behaviour tests:**
- `TestOpenAIBackend_DefaultBaseURL` — empty baseURL resolves to `api.openai.com`
- `TestOpenAIBackend_CustomBaseURL` — custom baseURL is used in the request URL
- `TestOpenAIBackend_TrailingSlashTrimmed` — trailing slash produces no `//chat/completions`

**Helm README:** AI Backend section updated — `openai` row now says "any OpenAI-compatible endpoint", new subsection with copy-paste examples for vLLM/LLMKube, Azure OpenAI, and Groq.

ntcdeveloper:
@Defilan PR's merged — opened the discussion to track the reference stack: https://github.com/wachd/wachd/discussions/59

Drop the verified config in there once you've stood it up on GPU hardware.

PR #65, opened 2026-06-05
feat(agent): wachd-agent with Kubescape CRD watcher

### LINKED ISSUE #64: feat(agent): wachd-agent — Kubescape CRD watcher, pluggable collector architecture
## Overview

Add `wachd-agent` — a lightweight binary deployed in customer clusters that watches security tool CRDs and forwards normalised events to the central Wachd instance for RCA, on-call routing, and escalation.

First release: Kubescape collector only. Falco and Radar collectors are future work.

## Architecture

One binary, pluggable collectors. Deployed once per customer cluster via a dedicated Helm chart.

```
wachd-agent (one Deployment per cluster)
├── collectors/
│   ├── kubescape/   ← CRD watcher (this issue)
│   ├── falco/       ← future
│   └── radar/       ← future
├── forwarder/       ← mTLS → central Wachd (written once, shared)
└── state/           ← last-seen dedup state, per collector
```

Each collector implements one interface:
```go
type Collector interface {
    Start(ctx context.Context) (<-chan Event, error)
}
```

## Kubescape collector — implementation spec

Based on architecture guidance from @matthyx (ARMO) in discussion #63.

### CRDs to watch
- `workloadconfigurationscansummaries` — config findings trigger
- `vulnerabilitymanifestsummaries` — vuln findings trigger
- One-shot GET on raw `workloadconfigurationscans` / `vulnerabilitymanifests` only when building alert body

### Reconcile loop
1. Watch summaries via aggregated apiserver (`spdx.softwarecomposition.kubescape.io/v1beta1`)
2. On `ADDED`/`MODIFIED` — skip if `metadata.generation` unchanged (no-op reconcile)
3. Gate on `kubescape.io/status == "completed"` before processing
4. Diff current severities against last-seen state:
   - Config: key on `controlID`
   - Vulns: key on `CVE-ID + PURL`
5. Fire only when `spec.severities.relevant.critical` or `.relevant.high` increases (vulns), or `spec.severities.critical`/`.high` increases (config)
6. One-shot GET raw manifest for CVE/control details → build alert body
7. Forward event to central Wachd via API token
8. Update last-seen state

### Routing labels (dedup key prefix)
- `kubescape.io/workload-namespace`
- `kubescape.io/workload-kind`
- `kubescape.io/workload-name`
- `kubescape.io/workload-container-name`
- `kubescape.io/image-id`

### RBAC footprint
```yaml
- apiGroups: ["spdx.softwarecomposition.kubescape.io"]
  resources:
    - workloadconfigurationscansummaries
    - vulnerabilitymanifestsummaries
    - workloadconfigurationscans
    - vulnerabilitymanifests
  verbs: ["get", "list", "watch"]
```

### Fail-open behaviour
- Watch errors: log, backoff, reconnect — never crash
- Forward errors: buffer locally, retry with exponential backoff
- If central Wachd is unreachable: keep watching, queue events, flush when connection restores

## New files

```
cmd/agent/
├── main.go                    ← binary entrypoint, config loading
├── collector.go               ← Collector interface
├── forwarder.go               ← mTLS/token forwarding to central Wachd
├── state.go                   ← last-seen dedup state (in-memory + optional persist)
└── collectors/
    └── kubescape/
        ├── watcher.go         ← CRD watch loop
        ├── reconciler.go      ← summary diff logic
        └── event.go           ← maps Kubescape CRD fields → wachd Event struct

helm/wachd-agent/
├── Chart.yaml
├── values.yaml
└── templates/
    ├── deployment.yaml
    ├── serviceaccount.yaml
    ├── clusterrole.yaml
    ├── clusterrolebinding.yaml
    └── secret.yaml            ← API token for central Wachd

Dockerfile.agent               ← distroless image, agent binary only
```

## CI changes

`ci.yml` — add second image build on `v*` tags:
```
ghcr.io/wachd/wachd:X.Y.Z        ← existing (server + worker + scheduler)
ghcr.io/wachd/wachd-agent:X.Y.Z  ← new (agent only, ~20MB)
```

## Acceptance criteria

- [ ] `wachd-agent` binary builds and runs standalone
- [ ] Kubescape collector watches `workloadconfigurationscansummaries` + `vulnerabilitymanifestsummaries`
- [ ] Only fires on net-new HIGH/CRITICAL findings (not re-scans of known findings)
- [ ] Gates on `kubescape.io/status == "completed"` before processing
- [ ] Uses `.relevant` severity for vuln findings
- [ ] Events forwarded to central Wachd and appear as incidents
- [ ] `helm/wachd-agent/` chart installs cleanly on a fresh cluster
- [ ] `ghcr.io/wachd/wachd-agent` image built and pushed by CI on `v*` tags
- [ ] Agent fails open — watch/forward errors never crash the process

## References

- Discussion #63 — full spec + Matthias's technical guidance
- Kubescape CRD types: https://github.com/kubescape/storage
- Annotation keys source: `kubescape/storage/tests/integration-test-suite/helpers.go`

ntcdeveloper:
## Summary

- Adds `wachd-agent`: a new binary that runs inside each customer cluster and watches Kubescape CRDs for net-new HIGH/CRITICAL security findings, then forwards them to the central Wachd webhook
- Kubescape is the first collector; the `agent.Collector` interface makes Falco, Radar, and others straightforward to add later
- Separate `Dockerfile.agent` and `helm/wachd-agent/` chart so customer clusters install only the tiny agent (~25MB distroless), never the full server image

### Architecture
- `internal/agent/event.go` — shared `Event` struct and `Collector` interface
- `cmd/agent/main.go` — entrypoint, fan-in goroutine, SIGTERM handling
- `cmd/agent/collectors/kubescape/collector.go` — watches `vulnerabilitymanifestsummaries` + `workloadconfigurationscansummaries`, fires only on net-new findings
- `cmd/agent/forwarder.go` — HTTP POST to `/api/v1/webhook/{teamID}/{secret}`
- `helm/wachd-agent/` — full Helm chart with ClusterRole (get/list/watch summaries only), hardened pod spec (nonroot, readOnlyRootFilesystem, drop ALL caps)
- CI: agent image built + Trivy-scanned alongside the main image on `v*` tags

### Key design decisions
- Uses `.relevant` severity for vulnerability findings (eBPF runtime signal — only CVEs in packages actually loaded at runtime), not `.all`. Cuts alert volume dramatically.
- `nestedInt` uses `NestedFieldNoCopy` + type switch to handle both `int64` (test objects) and `float64` (JSON decoder output from real k8s watch stream). The original `NestedInt64` would silently return 0 in production.
- Dedup state is in-memory per collector; on agent restart, collectors re-list existing objects to rebuild state before the watch stream begins.

## Test plan

- [x] `go test ./cmd/agent/... -race` — all pass
- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [ ] E2E: kind cluster + Kubescape install + vuln workload — see `docs/testing-wachd-agent.md`
- [ ] Verify RBAC: `kubectl auth can-i list pods --as system:serviceaccount:wachd-agent:wachd-agent` returns `no`
- [ ] Verify dedup: same finding count after re-scan produces no second event
- [ ] Verify `minSeverity=critical`: high-only workloads produce no event

Closes #64


fatkobra (changes requested):
**Request changes**
1. Agent payloads will likely be parsed as Grafana alerts
This forwarding path looks broken with the current server webhook parser.

The agent posts `{title,message,severity,source,labels}` to `/api/v1/webhook/{teamID}/{secret}`. But `parseWebhookPayload` checks Grafana before the generic fallback and classifies any JSON payload with `title` or `ruleName` as Grafana. The agent payload always has `title` and does not have Grafana `state`, so Kubescape findings will be stored as `source=grafana` with `severity=unknown`.

Can we add an explicit Wachd-agent/generic-source branch before Grafana detection, e.g. if `source` is present then use `title/message/severity/source`, or add a dedicated `/api/v1/agent/events` endpoint? Please add a server-side test that posts the agent payload and verifies `source=kubescape` and the original severity are preserved.

2. The collector does not prime state with an initial list before watching
The collector should prime its high-water state with an initial list before opening the watch.

The PR summary says the agent re-lists existing objects on restart to rebuild dedup state, but `runWatch` currently opens `Watch(ctx, metav1.ListOptions{})` directly. That leaves the state empty, so an existing completed summary can be emitted as new after deploy/restart, and reconnects have no `resourceVersion` continuity.

Can we switch this to the standard list-then-watch flow?

1. `List()` the summary GVR.
2. Reconcile/prime current completed objects without emitting.
3. Capture the list `metadata.resourceVersion`.
4. `Watch(... ListOptions{ResourceVersion: rv, AllowWatchBookmarks: true })`.
5. On expired resourceVersion / watch errors, re-list and restart.

A dynamic informer/Reflector would also solve this cleanly.

3. The generation gate is a fragile extra dedup layer
Can we avoid using `metadata.generation` as the first dedup gate?

The finding-count high-water check is the behavior we actually care about. If Kubescape updates summary output/status without changing generation, this early return can drop real new HIGH/CRITICAL totals before the count check runs.

I’d either remove the generation gate and dedupe only on the per-workload severity counts, or replace it with a scan identity/resourceVersion-based guard that is explicitly tied to Kubescape’s CRD contract.

**Non-blocking comments**
4. RBAC grants raw CRD access, but the collector does not fetch raw objects yet
Non-blocking: the chart grants `get` on the raw Kubescape CRDs, but the collector currently only watches summary CRDs and does not populate `Event.Details`.

For least privilege, can we either remove the raw CRD `get` permission until detail-fetching lands, or implement the raw-object fetch in this PR so the permission is justified?

5. Helm install should fail fast or support an existing secret
Could the chart either fail fast on missing required Wachd config or support `existingSecret`?

Right now `wachd.endpoint`, `teamID`, and `webhookSecret` default to empty, while `secret.yaml` always creates `<fullname>-secret` from `.Values.wachd.webhookSecret` and the Deployment always references that generated secret. The values comment says users can create the Secret manually, but there is no `existingSecret` escape hatch.

Suggested options:
- use Helm `required` for endpoint/teamID/webhookSecret, or
- add `wachd.existingSecret` / `existingSecretKey` and only render `secret.yaml` when it is not set.

ntcdeveloper:
All 5 addressed in the latest commit — here's what was done and where I'd welcome your take:

**1. Webhook parser misclassification**
Added a `source`-field check before the Grafana branch. If the payload has a non-empty `source`, we use it directly — agent payloads now come through as `source=kubescape` with the correct severity. The generic fallback was cleaned up to reuse the single `json.Unmarshal` rather than re-declaring `raw`.

Your review mentioned a dedicated `/api/v1/agent/events` endpoint as an alternative. I went with the source-field approach to avoid a new server-side route for now. Happy to switch if you think the dedicated endpoint is cleaner — it would let us enforce a stricter schema on agent payloads.

**2. List-then-watch**
Added `List()` on startup to prime dedup state, captures the `ResourceVersion`, then opens `Watch` with that rv and `AllowWatchBookmarks: true`. On any watch error the reconnect loop re-lists before re-watching, so no gap.

I used the manual list-then-watch pattern rather than a `Reflector`/`SharedInformer`. For 2 GVRs with simple reconcile logic it felt like the right weight. If you'd prefer a `Reflector` here I can switch — it would handle expired-rv backoff more robustly.

**3. Generation gate**
Removed entirely. Finding-count high-water mark is the only dedup now. Existing tests still pass since the count check was always the behaviour we cared about.

**4. RBAC**
Removed the raw CRD `get` rule. Only `get/list/watch` on summary CRDs remains.

**5. Helm**
Added `required` for `endpoint`/`teamID`/`webhookSecret`, and added `existingSecret` as an escape hatch — when set, `secret.yaml` is skipped and the deployment references the user-provided Secret directly.

fatkobra:
Fantastic job.
I do not see a blocking issue left from the code I checked.

fatkobra (approved):
I do not see a blocking issue left from the code I checked.

Issue #52, opened 2026-05-24
feat(graph): incident knowledge graph — full roadmap and missing pieces


ntcdeveloper:
## Overview

The incident knowledge graph was designed in #35 and the foundation has been built across multiple PRs. This umbrella issue maps everything that is done and what remains.

---

## What is built and shipped

| Work | PR | Status |
|---|---|---|
| Graph schema (`graph_nodes`, `graph_edges`), Store interface, node/edge type constants | #39 | ✅ Merged |
| `PostgresStore` implementation — `UpsertNode`, `PromoteNode`, `FindSimilar` (metadata scoring) | #40 | ✅ Merged |
| Worker writes resolved incident as permanent node on resolution | #50 | ✅ Merged |
| REST endpoints — `GET /graph/similar`, `GET /graph/subgraph` | #54 | ✅ Merged |
| Worker attaches similar past incident to alert notification | #55 | ✅ Merged |
| Web UI — similar incidents panel on incident detail page + graph settings | #56 | ✅ Merged |
| Service dependency graph for cross-service root cause analysis | #27 | ✅ Merged |
| Incident timeline view — chronological event log within a single incident | #57 | ✅ Merged |

---

## Remaining work

| Issue | What |
|---|---|
| #60 | Pending node write on alert fire + `similar_to` / `caused_by` edge writing on resolution |
| #61 | Service node registry — team self-service service declaration + `affects` edges (depends on #60) |
| #62 | Visual graph explorer on incident detail page (depends on #60) |

**Build order:** #60 → #61 and #62 in parallel (both depend on #60, not on each other)

---

## Out of scope for now

- pgvector / embedding-based similarity (current scoring is deterministic metadata matching)
- Cross-team graph (each team's graph is isolated by design)
- Automatic service discovery from traces (service dependencies are user-declared, not inferred)

ntcdeveloper:
All remaining work is shipped: #60 (pending node + edges), #61 (service registry), #62 (visual graph explorer, merged in #75). Closing umbrella.

PR #66, opened 2026-06-07
feat(graph): write pending nodes on alert fire and edges on resolution

### LINKED ISSUE #3: feat: Config-as-Code — wachd.yaml + wachd CLI

jothilal22:
## Summary

`graph_edges` was always empty in production ~ the store and `UpsertEdge` 
were implemented but never called. This PR adds the two missing write sites.

**Pending node on alert fire**  
A `pending` node is now written the moment an alert fires, before AI analysis 
runs. This gives the incident a graph presence from the start so `FindSimilar` 
can match against it while the incident is still active.

**Edges on resolution**  
After a node is promoted to permanent, two edges are now written:
- `similar_to` — links to past incidents with similar symptoms
- `caused_by` — links to the deployment that caused it, when AI identifies one

To support `caused_by` cleanly, added `IsDeploymentCause bool` to 
`AnalysisResponse` with a `DEPLOYMENT CAUSE: YES/NO` field in the AI prompt — 
no free-text grepping.

All graph writes are fail-open: errors are logged, alert routing is never blocked.

## Type of Change

- [x] New feature

## Testing

`make test` — all 15 packages pass, no external dependencies needed.

- 11 new unit tests covering both write paths, threshold filtering, 
  fail-open behaviour, and nil guards
- 3 new integration tests (auto-skip without DB, run in CI) confirming 
  actual rows in `graph_nodes` and `graph_edges`

## Related Issues

Closes #3  
Part of #52


ntcdeveloper (commented):
Reviewed the full diff. Solid approach overall — fail-open pattern is correct throughout, test coverage is good (11 unit + 3 integration), and the two-phase write (pending on fire → permanent on resolve) is the right design.

One blocking issue and two minor callouts.

---

## Blocking — `edgeSimilarThreshold` is redundant and masks store configuration

`writeSimilarEdges` guards at `score < edgeSimilarThreshold` (0.12), but `FindSimilar` already applies `s.minSimilarityScore` before returning results — see `postgres_store.go:382`:

```go
if match.Score < s.minSimilarityScore {
    continue
}
```

The default is `defaultMinimumSimilarityScore = 0.12`, so under the default config the worker check is dead code. Worse: if the store is instantiated with a lower threshold (via `WithMinimumSimilarityScore`), `FindSimilar` returns matches at e.g. 0.05+, but the worker silently discards everything below 0.12. The store's configured threshold is ignored.

Fix: remove `edgeSimilarThreshold` from the worker entirely and drop the `score < edgeSimilarThreshold` guard in `writeSimilarEdges`. `FindSimilar` is the authority on what clears the bar. `TestWriteIncidentEdges_SkipsBelowThreshold` should instead set `findSimilarNodes` to an empty slice (simulating the store already filtering), not rely on a worker-side guard.

---

## Minor — `Commits[0]` ordering assumption

`writeCausedByEdge` picks `sanitizedCtx.Commits[0]` as "the most recent commit". A short comment citing the collector's ordering contract (index 0 = newest) would prevent a future contributor from adding a collector path that returns commits in a different order and silently blaming the wrong commit.

---

## Minor — `writeIncidentEdges` return value

The function always returns nil (errors are logged internally). The `writeEdges func(ctx context.Context, nodeID uuid.UUID) error` signature in `persistResolvedIncidentNode` implies real errors may propagate. Worth a comment on `writeIncidentEdges` noting errors are logged and nil is always returned, so the intent is clear.

---

Everything else looks good. Checked `GetSubgraph` — the bidirectional SQL traversal (`e.from_id = walk.node_id OR e.to_id = walk.node_id`) means directed `similar_to` edges work correctly for the similar incidents panel without needing duplicate edges.

Adding @fatkobra for a second look.

fatkobra (commented):
## Non-blocking comments

### 1. Consider making the similarity threshold a single source of truth

```md
Non-blocking: can we avoid duplicating the similarity threshold here?

`FindSimilar` already applies the store’s `minSimilarityScore`, which defaults to `0.12`, and this PR adds a second hardcoded `edgeSimilarThreshold = 0.12` in the worker. That’s harmless today, but it can drift if the store threshold changes or becomes team-configurable later.

I’d either trust `FindSimilar` and remove the second check, or expose/use the same threshold constant from the graph package.
```

### 2. `caused_by` currently assumes `Commits[0]` is the root deployment


```md
Non-blocking: the `caused_by` edge currently links to `sanitizedCtx.Commits[0]` whenever `IsDeploymentCause=true`.

That’s probably fine for the first version if commits are guaranteed newest-first, but it means the AI can only say “a deployment caused this,” not which deployment. Could we add a short comment/test documenting the ordering guarantee, or track a follow-up to have the analysis return the causal SHA explicitly?
```

### 3. Nice-to-have: include properties on the pending node


```md
Nice-to-have/follow-up: pending nodes are currently label-only.

That’s enough for the current label-based similarity path, but if graph similarity starts considering properties later, active incidents won’t have service/source/severity context until resolution. No need to block this PR, but we may want to add lightweight pending properties such as severity, source, and service when available.
```

jothilal22:
@ntcdeveloper @fatkobra Fixed ~ removed edgeSimilarThreshold and the guard in writeSimilarEdges. FindSimilar is now the sole authority on the threshold. Updated the test accordingly. Also added the two comments on Commits[0] ordering and writeIncidentEdges return value.

ntcdeveloper (approved):
Fix looks correct —  removed, test updated, comments added. Good to merge after fatkobra signs off.

ntcdeveloper (approved):
Fix looks correct — edgeSimilarThreshold removed, test updated, comments added. Good to merge after fatkobra signs off.

fatkobra:
Great job @jothilal22 , I approve. Well done.

ntcdeveloper:
@jothilal22 Great first contribution — solid fix and clean tests, glad to have you here.

If you want to stay closer to the project, come join us on Discord: https://discord.gg/XAgXgnqxa — that's where we discuss roadmap, upcoming issues, and early features. Or drop us an email at sales@wachd.io.

Also, if any of your colleagues are into DevOps tooling or platform engineering, we'd love more contributors and a star always helps the project reach more people. 🙌

jothilal22:
@ntcdeveloper Thank you! I appreciate the feedback and the opportunity to contribute. I'll join the Discord and would be happy to stay involved with the project and help with upcoming issues as well.

jothilal22:
@ntcdeveloper Can you provide the valid discord link. Its say invalid link

ntcdeveloper:
here the valid one Sorr: https://discord.gg/GgQwzDkdC

Issue #60, opened 2026-06-04
feat(graph): write pending node on alert fire and edges on resolution


ntcdeveloper:
## Context

Part of the incident knowledge graph (#52). The store interface and `UpsertEdge` implementation exist but are never called from production code. `graph_edges` is always empty.

Two missing writes:

### 1. Pending node on alert fire
`writeResolvedIncidentToGraph` in `cmd/worker/graph.go` only writes on resolution. A `pending` node must also be written when the alert first fires so that `FindSimilar` can run while the incident is active (not just after it closes).

**Where to add it:** the alert processing path in `cmd/worker/`, after the incident record is created, before AI analysis runs.

### 2. Edge writing on resolution
When `writeResolvedIncidentToGraph` promotes a node to `permanent`, it should also write:
- `similar_to` edges — for each result returned by `FindSimilar` above threshold, call `UpsertEdge` with `EdgeTypeSimilarTo`
- `caused_by` edge — if a recent deployment is identified as probable cause in the AI analysis, write `EdgeTypeCausedBy` to the commit node

**Where to add it:** `persistResolvedIncidentNode` in `cmd/worker/graph.go`, after `PromoteNode` succeeds.

## Acceptance criteria

- [ ] A `pending` graph node is written when an alert fires
- [ ] `graph_edges` contains `similar_to` rows after two similar incidents resolve
- [ ] `graph_edges` contains a `caused_by` row when a deployment is identified as the cause
- [ ] All writes are team-scoped (`team_id` on every row)
- [ ] Fails open — graph write errors are logged and do not block alert routing
- [ ] Tests cover both write paths

## Notes

- `UpsertEdge` signature: `(ctx, teamID, *graph.Edge) (*graph.Edge, error)` — already implemented in `internal/graph/postgres_store.go`
- Edge type constants: `EdgeTypeSimilarTo`, `EdgeTypeCausedBy`, `EdgeTypeAffects` — defined in `internal/graph/store.go`
- Part of #52

ntcdeveloper:
Completed across two PRs:
- #66 — pending node on alert fire, `similar_to` and `caused_by` edges on resolution
- #70 — `affects` edge from resolved incident node to declared service node

All acceptance criteria met.

PR #70, opened 2026-06-09
feat(graph): add service node registry

### LINKED ISSUE #61: feat(graph): service node registry — team self-service service declaration
## Context

Part of the incident knowledge graph (#52). `NodeTypeService` exists as a constant in `internal/graph/store.go` but nothing ever creates service nodes. Without service nodes, `affects` and `depends_on` edges have no target to point at.

### API
Team self-service endpoints (admin role required):

```
GET    /api/v1/teams/{teamId}/graph/services
POST   /api/v1/teams/{teamId}/graph/services
DELETE /api/v1/teams/{teamId}/graph/services/{nodeId}
```

POST body:
```json
{
  "name": "checkout-api",
  "label": "Checkout API",
  "description": "Handles payment checkout flow"
}
```

Each service maps to a `graph_nodes` row with `type = 'service'` and `status = 'permanent'`.

### Edge writing
Once service nodes exist, the worker should write an `affects` edge (incident → service) when a resolved incident node is promoted. The service is matched by the `service` field extracted from the alert payload (`extractServiceName` in `cmd/worker/`).

### UI
A "Services" section in the Graph settings tab (already exists in `web/app/settings/page.tsx` via `graph-settings-panel.tsx`):
- List declared services
- Add service form (name + label)
- Delete button per service (admin only)

## Acceptance criteria

- [ ] Team admin can declare a service via POST and it appears in GET
- [ ] Declared services appear as `type=service, status=permanent` nodes in `graph_nodes`
- [ ] Worker writes an `affects` edge when a resolved incident's service name matches a declared service node
- [ ] All queries scoped by `team_id`
- [ ] UI shows service list in the Graph settings tab

## Depends on

#60 (edge writing infrastructure should be in place first)

## Notes

- Part of #52
- Service name matching uses the same `extractServiceName` logic already in the worker — no new extraction needed

fatkobra:
## Summary

Closes #61.

This PR adds a team-admin service registry backed by graph service nodes.

It adds:

- `GET /api/v1/teams/{teamId}/graph/services`
- `POST /api/v1/teams/{teamId}/graph/services`
- `DELETE /api/v1/teams/{teamId}/graph/services/{nodeId}`
- service nodes as `graph_nodes` rows with:
  - `type = service`
  - `status = permanent`
  - `external_id = normalized service name`
- a Services section in the Graph settings tab
- add/list/delete service UI
- worker `affects` edge writing from resolved incident node to declared service node

## Why

`NodeTypeService` already exists, but nothing creates service nodes yet. Without service nodes, `affects` edges have no stable target.

This PR lets team admins explicitly declare services like `checkout-api`, then resolved incidents can attach to those services when the incident service name matches.

## Implementation notes

- Service names are normalized by trimming and lowercasing before being stored as graph node `external_id`.
- The display label can be different from the service name.
- Description is stored in graph node properties.
- Worker service matching uses the existing `extractServiceName` output and the same normalization.
- If no matching service node exists, the worker skips the `affects` edge without failing the graph write.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation

## Testing

Added tests for:

- service node response mapping
- worker writes an `affects` edge when a declared service exists
- worker skips the `affects` edge when no declared service exists
- Graph settings Services section renders declared services
- Graph settings Services section creates a service
- Graph settings Services section removes a service after delete

Commands run:

```bash
go test ./cmd/server
go test ./cmd/worker
go test ./...
cd web
npm run test
npx eslint components/graph-services-section.tsx components/graph-services-section.test.tsx components/graph-settings-panel.tsx lib/api.ts lib/types.ts

## Code Review Checklist

## Related Issues

Closes #61.

### CODE REVIEW

**Comment by fatkobra** [comment]
Requesting review and good feedback here: @ntcdeveloper  @jothilal22

**Comment by jothilal22** [comment]
@fatkobra ~ Just thinking out loud after going through the PR.

### Blocking — normalization should be a shared function

The PR mentions that service names are normalized in the API handler (when storing) and that the worker applies the same normalization logic during lookup.

My concern is that if these are implemented as separate string operations in different places, they will eventually drift. For example, someone may update one side to replace spaces or handle additional cases and forget to update the other. At that point, `FindNodeByExternalID` starts silently missing matches.

I'd strongly recommend moving this into a single shared function (e.g. `normalizeServiceName`) and using it from both the API handler and `writeAffectsEdge`. That gives us one source of truth and avoids subtle bugs later.

### Blocking — handle empty service names before lookup

`extractServiceName` can return an empty string when a service cannot be determined.

In that case, we probably shouldn't call `FindNodeByExternalID` at all. Otherwise every resolved incident without a service tag ends up generating a `NodeNotFound` lookup and warning log, which can create a lot of noise over time.

I'd suggest an explicit early return:

```go
if serviceName == "" {
    return
}
```

The PR description mentions skipping affects-edge creation when no declared service exists, but I think the empty-string case should be handled explicitly as well.

### Minor — POST idempotency behavior should be documented

Since `UpsertNode` already handles duplicate `(team_id, type, external_id)` combinations, posting the same service multiple times shouldn't create duplicates.

It would be good to make the API behavior explicit though. For example:

* Does a repeated POST return `200 OK` with the existing service?
* Or does it return `409 Conflict`?

The current PR description doesn't mention this, and documenting it would make the API contract clearer.

### Minor — verify GET permissions

The PR adds a Services section under Graph Settings.

Assuming regular team members can already view graph-related information (incident panel, similar incidents, etc.), it probably makes sense for service listing to require team access rather than team admin permissions.

Worth double-checking that:

* `GET /graph/services` → team access
* `POST /graph/services` → team admin
* `DELETE /graph/services` → team admin

### Minor — keep it aligned with the existing edge-writing pattern

It would be nice if the affects-edge implementation followed the same fail-open pattern established in #60.

Something along the lines of:

```go
func writeAffectsEdge(...) {
    if serviceName == "" {
        return
    }

    // normalize
    // FindNodeByExternalID
    // UpsertEdge
    // log errors, never return them
}
```

And ideally it should be called from `writeIncidentEdges` alongside `writeSimilarEdges` and `writeCausedByEdge` rather than being wired separately. Keeping all edge creation logic in one place makes the flow easier to reason about and maintain.

Overall the test coverage looks good. I like that the PR covers the edge creation path, skip scenarios, UI CRUD flow, and includes eslint validation for the new components.

ntcdeveloper (commented):
Great work overall — the server handlers, ownership check, normalization, and tests all look solid. One bug before this can merge:

**`writeAffectsServiceEdge` is never called.**

`writeIncidentEdges` in `cmd/worker/graph.go` accepts `serviceName string` as a new parameter, but the function body only calls `writeSimilarEdges` and `writeCausedByEdge`. `writeAffectsServiceEdge` is never wired up, so the `affects` edge is never written despite the whole worker-side implementation being complete.

Go doesn't error on unused function parameters, so this compiles and tests pass — it would just silently do nothing in production.

Fix is a one-liner inside `writeIncidentEdges`:

```go
func writeIncidentEdges(ctx context.Context, graphStore graph.Store, teamID uuid.UUID, nodeID uuid.UUID, serviceName string, sanitizedCtx *correlator.Context, analysis *ai.AnalysisResponse) error {
    writeSimilarEdges(ctx, graphStore, teamID, nodeID)
    writeCausedByEdge(ctx, graphStore, teamID, nodeID, sanitizedCtx, analysis)
    writeAffectsServiceEdge(ctx, graphStore, teamID, nodeID, serviceName) // add this
    return nil
}
```

Everything else looks good — approve once this is added.

ntcdeveloper:
@jothilal22 good review — went through the same diff. A few notes on each point:

**Blocking #1 — shared normalization:** Valid. `normalizeGraphServiceName` (server) and `normalizeWorkerGraphServiceName` (worker) are identical one-liners right now, but any future change to one won't automatically apply to the other. Moving this into `internal/graph` or a small helper shared between both packages would be the right fix.

**Blocking #2 — empty service name guard:** This one is already handled. `writeAffectsServiceEdge` does `name := normalizeWorkerGraphServiceName(serviceName)` then immediately `if name == "" { return }` — so an empty string from `extractServiceName` exits before the DB lookup, no noise in the logs.

**Minor — GET permissions:** Confirmed. All three handlers use `requireTeamAdmin` right now. GET should be downgraded to team-member access — listing declared services is read-only config, same as viewing the graph. Only POST and DELETE need admin.

**Minor — POST idempotency + wire-up alignment:** Both valid, agree with the direction.

Also posted separately: the missing `writeAffectsServiceEdge` call in `writeIncidentEdges` — serviceName is threaded in but never forwarded to the function. That's the other blocking item before this can merge.

fatkobra:
Thanks for the reviews. I pushed a follow-up commit addressing the blocking items and the main minor points.

Changes made:

- moved service-name normalization into a shared `graph.NormalizeServiceName(...)` helper
- updated both the API service handlers and worker affects-edge lookup to use that shared helper
- added unit coverage for the shared normalizer
- wired `writeAffectsServiceEdge(...)` into `writeIncidentEdges(...)`
- kept the empty service-name guard explicit so incidents without a detected service do not call `FindNodeByExternalID`
- changed `GET /graph/services` to require team access instead of team admin
- kept `POST /graph/services` and `DELETE /graph/services/{nodeId}` admin-only
- documented POST idempotency in code:
  - new service returns `201 Created`
  - repeated POST for the same normalized service name updates the node and returns `200 OK`
- added a worker test proving an empty service name does not trigger lookup or edge creation
- added a worker test assertion proving lookup uses the normalized service name

Checks run:

```bash
go test ./internal/graph
go test ./cmd/server
go test ./cmd/worker
go test ./...
cd web
npm run test
npx eslint components/graph-services-section.tsx components/graph-services-section.test.tsx components/graph-settings-panel.tsx lib/api.ts lib/types.ts

ntcdeveloper (approved):
All blocking items addressed:

- `NormalizeServiceName` in `internal/graph` is the single source of truth — both API handler and worker lookup use it, normalization cannot drift
- `writeAffectsServiceEdge` is now wired into `writeIncidentEdges` alongside `writeSimilarEdges` and `writeCausedByEdge`
- Empty service name guard is explicit, no DB lookup noise on incidents without a detected service
- GET uses `requireTeamAccess`, POST/DELETE use `requireTeamAdmin`
- POST idempotency is clear: 201 new, 200 existing
- CI green

Approved.

ntcdeveloper:
@jothilal22 @fatkobra addressed everything from your review — shared normalizer in `internal/graph`, `writeAffectsServiceEdge` wired up, empty service name guard, GET permissions fixed, POST idempotency documented. I've approved. Would you mind doing a formal re-review when you get a chance?

jothilal22 (approved):
Good job ! Looks good to me ! Approved

ntcdeveloper:
@fatkobra both approvals are in and CI is green. The branch just needs a rebase on main since #71 landed — could you run `git rebase main` on your fork and push? No conflicts expected, the two PRs touch completely different files.

fatkobra:
Rebased onto the latest `main` after #71 landed.

I resolved the rebase conflicts, reran the relevant tests, and force-pushed the updated branch.

The PR should now be up to date and ready for merge.

@ntcdeveloper could you please take a look?

ntcdeveloper:
Merged manually after rebasing on main — squash commit 12a2ff7. Closes #61.

PR #71, opened 2026-06-10
feat(agent): add k8shealth collector — standalone pod/node alerting

### LINKED ISSUE #69: feat(agent): standalone alert generation — remove Grafana as a hard dependency
## Summary

Right now Wachd requires an external alerting tool (Grafana, Datadog, etc.) to fire the webhook before anything happens. Teams without a full observability stack are blocked at step one.

`wachd-agent` is already running inside the cluster with RBAC, a pluggable collector loop, and a connection back to Wachd central. It can become the alert source itself — no Grafana required.

### 1. Pod / node health watcher (zero dependencies)
- Watch k8s API for `CrashLoopBackOff`, `OOMKilled`, `ImagePullBackOff`, `Pending > N min`
- Watch node conditions: `NotReady`, `MemoryPressure`, `DiskPressure`
- No Prometheus, no Grafana — pure k8s API
- Highest value for bare clusters

### 2. Prometheus rule evaluator
- Agent queries Prometheus `/api/v1/query` directly
- Team configures PromQL rules + thresholds in Wachd UI
- Covers teams with Prometheus but no Grafana

### 3. Falco inbound listener
- Agent exposes a local HTTP endpoint
- Falco `http_output` points at it (no Falcosidekick needed)
- Agent normalises Falco payload → standard Wachd webhook format → Wachd central

### 4. Loki error rate (future)
- Query Loki API for error rate spikes
- For teams already running Loki without Grafana alerting configured

## Architecture rule

The agent stays a normaliser + forwarder only — no RCA, no scheduling, no AI. All signals are normalised to the standard Wachd webhook payload and POSTed to `/webhook/{teamId}/{secret}` on Wachd central. Intelligence stays in the server.

```
Cluster
├── Kubescape    → CRD watch          ──┐
├── Falco        → http_output        ──┤──→ wachd-agent → Wachd central → RCA + on-call
├── Prometheus   → PromQL eval        ──┤
└── k8s API      → pod/node watch     ──┘
     (no deps)
```

## Out-of-the-box integration support

Because the collector loop is already pluggable (Kubescape is collector #1), any new signal source — Falco, Prometheus, Trivy, custom tools — is just a new collector file. No changes to the agent core, no new RBAC, no new deployment.

## What this unlocks

| Team setup | Alert source today | Alert source after |
|---|---|---|
| Full stack (Grafana + Kubescape + Falco) | Grafana webhook | All three, unified |
| Prometheus only, no Grafana | ❌ Blocked | Agent evaluates PromQL rules |
| Kubescape only | CVE/compliance only | Same, already works |
| Bare cluster, nothing | ❌ Blocked | Agent pod/node health watcher |
| Falco only | ❌ Blocked | Agent receives Falco `http_output` |

## Positioning

> Drop one agent into any cluster. Wachd handles the rest.

No Grafana prerequisite. No observability stack prerequisite. Every team gets smart alerting from day one.

## Build order

1. Pod/node health watcher — highest impact, zero deps
2. Prometheus rule evaluator — covers the "Prometheus but no Grafana" gap
3. Falco inbound listener — Falco integration already discussed with Thomas Labarussias

ntcdeveloper:
## Summary

Adds the `k8shealth` collector to `wachd-agent`, eliminating the hard dependency on Grafana for basic Kubernetes health alerting.

- **Pod watcher**: fires on `CrashLoopBackOff`, `ImagePullBackOff`, `ErrImagePull`, and `OOMKilled` — deduplicated per active condition, re-fires on each distinct OOM (new `startedAt`)
- **Node watcher**: fires on `NotReady` (critical) and pressure conditions — `MemoryPressure`, `DiskPressure`, `PIDPressure` (high)
- **Pending poller**: polls every minute, fires when a pod is stuck in `Pending` longer than `K8SHEALTH_PENDING_MINUTES` (default: 15)
- **Recovery**: clears dedup state when a pod returns to `Running` with all containers ready — so a crash-recover-crash cycle fires on every crash, not just the first
- **Cluster label**: forwarder now injects optional `WACHD_CLUSTER` label on every outgoing event from all collectors (Kubescape + k8shealth)

Requires no external tools — uses the Kubernetes API directly via in-cluster config.

## Changes

- `cmd/agent/collectors/k8shealth/collector.go` — new collector (431 lines)
- `cmd/agent/collectors/k8shealth/collector_test.go` — 11 unit tests
- `cmd/agent/forwarder.go` — cluster label injection
- `cmd/agent/forwarder_test.go` — cluster label tests
- `cmd/agent/main.go` — wire up k8shealth + WACHD_CLUSTER env var
- `helm/wachd-agent/` — k8shealth values, deployment env vars, ClusterRole RBAC
- `docs/testing-wachd-agent.md` — E2E test steps for all new scenarios

## Test plan

- [x] Unit tests: `go test ./cmd/agent/...` — all pass
- [x] E2E (kind cluster): CrashLoopBackOff fires + deduplicates
- [x] E2E: OOMKilled fires separately per restart
- [x] E2E: stuck-Pending fires after threshold (2m in test)
- [x] E2E: NodeNotReady fires as critical
- [x] E2E: disabled collector (`K8SHEALTH_ENABLED=false`) emits nothing
- [x] E2E: cluster label present on all forwarded events
- [x] RBAC: only pods/nodes get/list/watch — no secrets, no write access

Closes #69


ntcdeveloper:
Hey @fatkobra and @jothilal22 — would love your eyes on this when you get a chance. This is the first slice of #69 (pod/node health watcher). The architecture is intentionally simple: detect → POST webhook → done. Worker handles the rest. The E2E steps are in `docs/testing-wachd-agent.md` if you want to spin it up locally.

jothilal22:
@ntcdeveloper  ~ Overall, this is a solid first slice. The scope feels right, the implementation is straightforward, and the dedup approach is well thought out. A couple of things stood out while reading through it.

### Blocking ~ recovery behavior needs to be explicit

The PR mentions that dedup state is cleared when a pod returns to `Running` with all containers ready. My question is whether that recovery also results in a `resolved` event being sent to central, or if it only clears local state.

If we're only clearing dedup state, incidents in Wachd will remain open indefinitely. Ideally, recovery should trigger a `status: resolved` event using the same fingerprint that was used for the firing event.

If that's already happening, it would be good to call it out explicitly in the PR description. If not, I think that's a missing piece before merge.

### Blocking ~ potential fan-out during node failures

We discussed this in #69, but I don't see it addressed in the current proposal.

If a node becomes `NotReady`, the node watcher will generate an alert. At the same time, the pods running on that node may start failing and generate pod-level alerts as well. In a busy cluster that could easily turn into dozens of incidents for what is effectively a single root cause.

It would be good to either:

* Suppress pod alerts when their assigned node is already known to be `NotReady`, or
* Call this out as a known limitation and create a follow-up issue

Otherwise the fan-out problem remains unresolved.

### Minor ~ RBAC scope

The ClusterRole looks minimal and sensible, but I wonder whether namespace-scoped mode should be the default.

Some teams are very sensitive about cluster-wide permissions, and requiring a ClusterRole from day one may become an adoption hurdle. Making cluster-wide monitoring an explicit opt-in via Helm values could be a nice middle ground.

Not a blocker, just something worth considering.

### Minor — dedup state is lost on restart

The dedup cache appears to be in-memory only.

That seems perfectly reasonable for v1, but it does mean that after an agent restart, any currently failing pods will likely generate alerts again until state is rebuilt. That's probably acceptable, but it would be helpful to document it so teams know what to expect during rollouts or node evictions.

### Minor — schema version

We raised this in #69 as well. Adding a simple:

```json
{
  "schema_version": "1"
}
```

to the normalized payload now is very cheap and gives us room to evolve the payload format later without compatibility headaches.

### What looks good

* Deduping `OOMKilled` events using `startedAt` is the right trade-off
* Treating `Pending` as a periodic poll instead of a watcher makes sense
* RBAC scope is refreshingly small (read-only pods/nodes, no secrets, no writes)
* Injecting the cluster label in the forwarder is the right abstraction layer
* Test coverage looks strong for a first collector
* `K8SHEALTH_ENABLED=false` is a useful operational escape hatch

For me, the two things worth resolving before merge are:

1. Making sure recovery generates resolved events.
2. Addressing or explicitly tracking the node failure fan-out scenario.

Everything else feels like reasonable follow-up work.

fatkobra:
Thanks for putting this together.

This is exactly the first standalone-agent slice I was hoping for: Wachd can start generating useful alerts from Kubernetes pod/node health without requiring Grafana, Prometheus, or another monitoring stack.

What looks good:

- `k8shealth` is isolated as its own collector.
- The collector covers the right first failure modes:
  - CrashLoopBackOff
  - ImagePullBackOff / ErrImagePull
  - OOMKilled
  - stuck Pending pods
  - node NotReady / Unknown
  - node pressure conditions
- The in-memory `seen` map prevents repeated alerts for the same active condition.
- Recovery clears state so a future crash can fire again.
- The cluster label injection in the forwarder is useful for multi-cluster installs.
- Helm RBAC is read-only and limited to the resources needed for this collector.
- Tests cover the important reconcile behavior.

I do have one blocking concern before merge.

### Blocking: run an initial list/reconcile before opening watches

Right now `runPodWatch` and `runNodeWatch` open Kubernetes watches directly.

That means the collector reacts to new watch events, but it may miss bad conditions that already exist before the agent starts.

Examples:

- a pod is already in `CrashLoopBackOff` before `wachd-agent` starts
- a pod is already in `ImagePullBackOff`
- a container was already `OOMKilled`
- a node is already `NotReady`

The Pending case is partly covered by the poller, but the pod waiting/terminated states and node conditions depend on watch events.

I think the collector should do an initial list/reconcile pass before entering the watch loop:

- list all pods across the configured scope
- call `reconcilePod(...)` for each
- emit any events that should fire
- list all nodes
- call `reconcileNode(...)` for each
- then start watching for changes

This can be done directly with a list-before-watch approach, or by moving to a shared informer later. For this PR, a direct initial list pass is probably enough.

I would also add tests for:

- existing CrashLoopBackOff pod is emitted during initial pod list
- existing NotReady node is emitted during initial node list

### Non-blocking: consider namespace scoping

The Helm RBAC is cluster-wide for pods and nodes, and the collector watches pods across all namespaces.

That is a reasonable default for a cluster-level agent, but I think it would be useful to support namespace scoping for pods in a follow-up:

```text
K8SHEALTH_NAMESPACES=default,payments,checkout

fatkobra (changes requested):
### Blocking: run an initial list/reconcile before opening watches

Right now `runPodWatch` and `runNodeWatch` open Kubernetes watches directly.

That means the collector reacts to new watch events, but it may miss bad conditions that already exist before the agent starts.

Examples:

- a pod is already in `CrashLoopBackOff` before `wachd-agent` starts
- a pod is already in `ImagePullBackOff`
- a container was already `OOMKilled`
- a node is already `NotReady`

The Pending case is partly covered by the poller, but the pod waiting/terminated states and node conditions depend on watch events.

I think the collector should do an initial list/reconcile pass before entering the watch loop:

- list all pods across the configured scope
- call `reconcilePod(...)` for each
- emit any events that should fire
- list all nodes
- call `reconcileNode(...)` for each
- then start watching for changes

This can be done directly with a list-before-watch approach, or by moving to a shared informer later. For this PR, a direct initial list pass is probably enough.

I would also add tests for:

- existing CrashLoopBackOff pod is emitted during initial pod list
- existing NotReady node is emitted during initial node list

ntcdeveloper:
Thanks both — good catches.

**@fatkobra — initial list/reconcile:** Agree, this is a real gap. A pod already in CrashLoopBackOff before the agent starts will be missed until a new watch event arrives. Will add an initial list pass before entering each watch loop and cover it with tests. Pushing a fix.

**@jothilal22 — resolved events:** Looked at the server side for this. `handleWebhook` always creates a new incident regardless of payload content — there's no resolved-event path in the webhook receiver today for any source. Adding resolved events from the agent requires both server-side changes (detect and auto-close an existing incident by fingerprint) and agent-side changes (emit a resolve payload on recovery). That's a meaningful separate feature. Opening a follow-up issue to track it properly.

**@jothilal22 — node failure fan-out:** Valid known limitation. Suppressing pod alerts when the node is already NotReady adds cross-watcher state and makes the logic more complex than this first slice warrants. Will document it in the PR and add it as a follow-up to #69.

**Other minor points** (dedup on restart, schema_version, namespace scoping) — all reasonable follow-up work, agree with the direction.

The only code change coming in this update is the initial list/reconcile fix.

ntcdeveloper:
Blocking this PR on #72.

@fatkobra @jothilal22 — while reviewing the resolved-events discussion above I went and checked the webhook receiver end-to-end. The dedup gap is deeper than just the agent:

- `handleWebhook` unconditionally inserts a new open incident on every POST — no fingerprint check, no existing-open check
- `parseWebhookPayload` does not return a resolved/firing flag
- The `incidents` table has no `fingerprint` column and no unique constraint on any alerting key
- Grafana resolved events (`state: ok`) have never actually closed an incident — they just open a new low-severity one

The k8shealth agent will make this significantly worse. A single flapping pod (crash → recovery → crash) creates a separate open incident on each crash cycle. A busy cluster could generate dozens of open incidents for the same pod within minutes.

#72 covers what needs to be built: fingerprint column, dedup on ingest, resolve on ingest, and `parseWebhookPayload` returning a status flag.

Holding #71 until #72 is merged. Happy to hear if either of you sees a reason to approach this differently.

fatkobra:
Agree with holding #71 on #72.

The collector is valuable, but the receiver currently has no incident identity or resolve-on-ingest path. That means k8shealth would turn pod flapping into many duplicate open incidents, even if the agent-side dedup is mostly correct.

Suggested path:

1. Land #72 first:
   - add incident fingerprint
   - dedup firing alerts on ingest
   - resolve matching open incidents on resolved events
   - make parseWebhookPayload return firing/resolved status
   - add regression tests for duplicate firing and Grafana state=ok

2. Rebase #71 after #72.

3. Then finish #71-specific review items separately.

ntcdeveloper:
Rebased on main — PR #73 (incident dedup + fingerprint) is now in the base. The one remaining blocking item is @fatkobra's initial list/reconcile fix: pods already in a bad state before the agent starts need to be listed once before the watch loop opens.

ntcdeveloper:
Just pushed the last piece — initial list/reconcile fix so pods and nodes already in a bad state before the agent starts are caught immediately, not only on the next watch event.

- `runPodWatch` and `runNodeWatch` now do a `List` first, reconcile each object, then open `Watch` from that `ResourceVersion` (standard k8s list+watch pattern)
- Added `TestStart_InitialPodReconcile` and `TestStart_InitialNodeReconcile` using the fake k8s client to cover the gap
- All 13 tests pass, clean build

@fatkobra @jothilal22 ready for another look 🙏

jothilal22 (approved):
@ntcdeveloper looks good to me !! Cheers

fatkobra (approved):
Great work, this looks good to go.
Aproved.

PR #73, opened 2026-06-10
fix(server): incident dedup and resolve via fingerprint

### LINKED ISSUE #72: fix(server): incident dedup and resolve-on-ingest at webhook receiver
## Problem

Every call to `POST /api/v1/webhook/{teamId}/{secret}` unconditionally inserts a new open incident, regardless of whether an identical alert is already open.

There is no dedup or resolve logic at any layer:

- `handleWebhook` has no fingerprint check — always calls `db.CreateIncident(...)`
- `parseWebhookPayload` does not return a resolved/firing status
- The `incidents` table has no `fingerprint` column and no `UNIQUE` constraint on any alerting key
- The worker has no dedup logic
- The store has no upsert or `ON CONFLICT` for incidents

### What this means in practice

**Duplicate open incidents:** If Grafana fires the same alert twice, or the k8shealth agent fires CrashLoopBackOff → recovery → CrashLoopBackOff, the server creates 2, 3, 4 separate open incidents for the same condition. The on-call engineer gets paged multiple times for one problem.

**Resolved events are silently broken:** Grafana sends `state: "ok"` on recovery. The server detects this (sets severity to `"low"`) but still calls `db.CreateIncident(...)`. The resolved event opens a new low-severity incident instead of closing the original. This has never worked.

### Why this blocks the k8shealth agent (PR #71)

PR #71 adds a pod/node health collector that can fire CrashLoopBackOff alerts. Pods crash and recover. Without dedup, each crash cycle after a recovery creates another open incident. A single flapping pod in a busy cluster can generate dozens of open incidents within minutes.

Merging the agent before this is fixed makes the problem significantly worse.

### 1. Add `fingerprint` to the incidents table

```sql
ALTER TABLE incidents ADD COLUMN IF NOT EXISTS fingerprint CHAR(64);
CREATE INDEX IF NOT EXISTS idx_incidents_fingerprint ON incidents(team_id, fingerprint);
```

Fingerprint = SHA-256 of `team_id + source + normalized_title`.

### 2. `handleWebhook` — dedup on ingest

Before inserting, check for an existing open incident with the same fingerprint:

```go
existing, err := s.db.FindOpenIncidentByFingerprint(ctx, teamID, fingerprint)
if existing != nil {
    // already open — return 202, no duplicate
    w.WriteHeader(http.StatusAccepted)
    return
}
```

### 3. `handleWebhook` — resolve on ingest

If the incoming payload signals `status: "resolved"` (Grafana `state: "ok"`, agent future resolve events), close the matching open incident instead of creating a new one:

```go
if isResolved {
    s.db.ResolveIncidentByFingerprint(ctx, teamID, fingerprint)
    w.WriteHeader(http.StatusAccepted)
    return
}
```

### 4. `parseWebhookPayload` — return firing/resolved status

The function currently returns `(title, message, severity, source)`. It needs to also return `isResolved bool` so the caller can branch on it.

## Scope

This is a server-only change. The agent and worker are not affected. The k8shealth agent PR (#71) should be held until this is merged.

## Related

- Blocks #71 — k8shealth agent will flood open incidents without this
- PR #66 — similarity matching works at graph layer after incidents are stored; this is the missing upstream dedup layer

ntcdeveloper:
## Problem

Every webhook POST created a new open incident — no dedup, no resolve path. A Grafana `state: "ok"` callback was silently creating a new `low`/`unknown` severity incident instead of closing the original. Same for wachd-agent `status: "resolved"` and Datadog `alert_type: "success"`.

This means the similarity search, incident graph, and on-call escalation were all built on top of a store that could have hundreds of duplicate incidents for a single real event.

Closes #72.
Unblocks #71.

## What changed

**Schema** (`internal/store/schema.sql`):
- Added `fingerprint CHAR(64)` column with a partial index on `(team_id, fingerprint)` where not null

**Store** (`internal/store/`):
- `Incident` struct: `Fingerprint *string` field
- `CreateIncident`: includes fingerprint in INSERT
- `FindOpenIncidentByFingerprint`: returns most-recent open/acked/snoozed incident for a fingerprint, nil if none
- `ResolveIncidentByFingerprint`: UPDATE to resolved where status IN (open/acknowledged/snoozed); no-op if already resolved

**Webhook handler** (`cmd/server/main.go`):
- `parseWebhookPayload` now returns a 5th `resolved bool` — Grafana `state=ok`, generic `status=resolved`, Datadog `alert_type=success`
- `incidentFingerprint(teamID, source, title)` — SHA-256 of normalized title (lower-cased, trimmed)
- `handleWebhook` flow:
  1. If resolved → `ResolveIncidentByFingerprint` → 202
  2. If open duplicate exists → 202 deduplicated (returns existing incident_id)
  3. Otherwise → create incident as before
  4. Fingerprint lookup errors are logged but do not drop the alert (fail-open)

## Tests

7 new unit tests in `webhook_parser_test.go`:
- `TestParseWebhookPayload_GrafanaResolvedReturnsTrue`
- `TestParseWebhookPayload_AgentResolvedReturnsTrue`
- `TestParseWebhookPayload_DatadogResolvedReturnsTrue`
- `TestIncidentFingerprint_Deterministic`
- `TestIncidentFingerprint_NormalizesTitle`
- `TestIncidentFingerprint_DifferentTeamsDiffer`
- `TestIncidentFingerprint_DifferentSourcesDiffer`

All existing tests updated for the new 5-value return signature.

## Manual verification

Tested end-to-end against a live server + Postgres:

```
POST firing alert     → 200 {"status":"accepted","incident_id":"3bd606c7-..."}
POST same alert again → 202 {"status":"deduplicated","incident_id":"3bd606c7-..."}
POST resolved event   → 202 {"status":"resolved"}
DB: 1 row, status=resolved, fingerprint=a4188e0c...
```

## Checklist

- [x] All database queries scoped by `team_id`
- [x] Parameterized queries only (no string concat)
- [x] Schema migration uses `IF NOT EXISTS` (safe to re-run)
- [x] Fail-open: fingerprint errors never drop an alert
- [x] `go test ./...` all green
- [x] `go build ./...` clean

### DIFF
--- a/cmd/server/e2e_integration_test.go
+++ b/cmd/server/e2e_integration_test.go
@@ -43,6 +43,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"sync"
 	"testing"
 	"time"
 
@@ -62,6 +63,8 @@ import (
 type e2eEnv struct {
 	db                  *store.DB
 	router              *mux.Router
+	server              *Server
+	q                   *queue.Queue
 	kongTeam            *store.Team
 	kafkaTeam           *store.Team
 	kongCookie          *http.Cookie // admin session scoped to Kong team only
@@ -186,6 +189,8 @@ func newE2EEnv(t *testing.T) *e2eEnv {
 	return &e2eEnv{
 		db:                  db,
 		router:              router,
+		server:              server,
+		q:                   q,
 		kongTeam:            kongTeam,
 		kafkaTeam:           kafkaTeam,
 		kongCookie:          kongCookie,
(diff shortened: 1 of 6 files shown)


ntcdeveloper:
@fatkobra @jothilal22 — PR #73 is ready for review.

This fixes the dedup gap we discussed on #71 and #72. Every webhook POST was unconditionally creating a new incident — no fingerprint, no dedup, no resolve path. Grafana `state: "ok"` callbacks were silently opening new low-severity incidents instead of closing the original.

**What's in this PR:**
- `fingerprint CHAR(64)` column on incidents (SHA-256 of `teamID|source|normalized_title`)
- `FindOpenIncidentByFingerprint` — dedup before create
- `ResolveIncidentByFingerprint` — close on resolved webhook (Grafana `state=ok`, agent `status=resolved`, Datadog `alert_type=success`)
- Fail-open: fingerprint errors never drop an alert
- 7 new unit tests + all existing tests updated

E2E verified locally: fire → dedup → resolve → 1 row in DB, status=resolved.

Once this merges I'll rebase #71 on main.

fatkobra (changes requested):
Thanks for jumping on this quickly. This is definitely the right fix direction for #72 and it should unblock #71 once it is safe enough to rely on.

I am going to keep this blocked for now because I think the current implementation still leaves a few important correctness gaps in the webhook receiver.

### Blocking

1. **Dedup is still race-prone**

Right now the flow is:

```go
existing, err := s.db.FindOpenIncidentByFingerprint(...)
if existing == nil {
    s.db.CreateIncident(...)
}
```
but the schema only adds a non-unique index on (team_id, fingerprint). Two identical webhook POSTs arriving at the same time can both miss the lookup and both insert open incidents.

For this fix to actually close the dedup gap, the database needs to enforce the invariant too. Suggested approach:
```SQL
CREATE UNIQUE INDEX IF NOT EXISTS idx_incidents_active_fingerprint_unique
ON incidents(team_id, fingerprint)
WHERE fingerprint IS NOT NULL
  AND status IN ('open', 'acknowledged', 'snoozed');
```
Then make the create path handle the conflict by returning the existing incident id / dedup response. A transaction/advisory lock around team_id + fingerprint would also work, but I would prefer DB-backed uniqueness here.

2. **Resolved and duplicate events are still blocked by the monthly alert limit**

handleWebhook currently checks CountIncidentsThisMonth before reading/parsing the payload. That means once the tenant hits the monthly limit, a resolved webhook gets 429 and cannot close the existing incident. Duplicate firing alerts also get rejected instead of deduplicated.

The quota check should only apply when we are about to create a new incident. Suggested order:

- auth/rate-limit
- read body
- parse payload
- compute fingerprint
- if resolved -> resolve and return 202
- if duplicate active incident -> return 202 deduplicated
- check monthly alert limit
- create new incident
- enqueue alert job

3. **Webhook resolve should trigger the same post-resolve flow as manual resolve**

The manual resolve endpoint updates the incident and enqueues `EnqueueIncidentResolved(...)`. The new webhook resolve path only updates the DB through `ResolveIncidentByFingerprint(...)` and returns `202`.

Please make the fingerprint resolve path return the resolved incident id(s), then enqueue the same resolved job for each affected incident. Otherwise webhook-based recovery can bypass graph/worker/resolution side effects that manual resolution already triggers.

4. **Please add regression coverage for the actual ingest/store behavior**

The new parser/fingerprint tests are useful, but #72 is about receiver behavior. I would like to see tests covering the end-to-end server/store cases, not just parseWebhookPayload.

Minimum cases I would expect before merge:

- first firing webhook creates one incident with fingerprint
- duplicate firing webhook returns deduplicated and does not create a second row
- resolved webhook closes the existing open incident and does not create a low-severity incident
- resolved webhook still works when the monthly create limit has been reached
- concurrent duplicate firing webhooks cannot create two active incidents for the same fingerprint

5. **Clean up the hidden/bidi Unicode warning**

There are warnings about hidden/bidirectional Unicode in changed files. Please remove those characters or confirm they pre-existed and are not part of this PR. Useful check:
```Bash
git grep -nP '[\x{202A}-\x{202E}\x{2066}-\x{2069}]'
gofmt -w cmd/server/main.go cmd/server/webhook_parser_test.go internal/store/*.go
```
### Non-blocking / follow-up

It would also be good to normalize incoming status fields before comparison:
```Go
strings.ToLower(strings.TrimSpace(status))
```
for Grafana state, generic status, and Datadog alert_type / alert_transition.

Also, longer term, team_id + source + normalized_title is okay for this first fix because that is what #72 asked for, but we should probably add explicit upstream fingerprints later when available, especially for Grafana/Datadog alerts with the same title but different labels/resources.

Once the atomic dedup, quota ordering, resolve side effects, and regression tests are fixed, I am happy to re-review quickly. The PR is close, but I do not think #72 is fully closed yet.

ntcdeveloper:
All 5 blocking items addressed in the latest commit (`0c9de9d`):

**1. Race-safe dedup**
Added `UNIQUE` partial index `idx_incidents_active_fingerprint_unique` on `(team_id, fingerprint)` WHERE `status IN ('open','acknowledged','snoozed')`. `CreateIncident` now detects `23505` unique_violation and returns `store.ErrDuplicateIncident`. `handleWebhook` catches it, looks up the winner, and returns deduplicated — so concurrent identical POSTs can never create two active incidents.

**2. Quota check order**
Body read, parse, fingerprint compute, resolve path, and dedup path all run before `CountIncidentsThisMonth`. Resolved and duplicate webhooks are never blocked by the monthly limit. The quota check now only gates actual new incident creation.

**3. Resolve side effects**
Webhook resolve path now calls `FindOpenIncidentByFingerprint` before resolving, then enqueues `EnqueueIncidentResolved` for the closed incident — same post-resolve flow as the manual endpoint.

**4. Integration tests**
`TestWebhookDedup` in `e2e_integration_test.go` (build tag: `integration`) covers all five requested cases:
- `first_fire_creates_incident_with_fingerprint` — 202/accepted + 64-char fingerprint in DB
- `duplicate_returns_deduplicated_no_second_row` — same incident_id, exactly 1 DB row
- `resolve_closes_incident_no_new_row` — status=resolved in DB, still 1 row
- `resolve_works_at_quota_limit` — new alerts blocked (429), resolve still gets 202
- `concurrent_duplicates_create_one_incident` — 10 goroutines fired simultaneously, 1 DB row

All 5 pass against real Postgres + Redis.

**5. Bidi Unicode / formatting**
`git grep` found no hidden Unicode. `gofmt -w` applied to all changed files — clean.

**Non-blocking (also done):** normalized `status`, `state`, and `alert_type` with `strings.ToLower(strings.TrimSpace(...))` before comparison across all three webhook sources.

ntcdeveloper:
@jothilal22 — sent you a collaborator invite. Once you accept, I'll add you as a reviewer. Happy to have your eyes on this one given your feedback on #71.

fatkobra (approved):
Great job with this.
I approve

jothilal22:
Accepted the invite.

jothilal22 (approved):
Great work @ntcdeveloper Looks good to me !!

PR #74, opened 2026-06-11
fix: resolve-on-ingest for generic webhooks + affects/similar_to edges

ntcdeveloper:
## Summary

Three bugs found during graph feature testing (issues #60/#61) that blocked the entire resolve-on-ingest path for generic webhook payloads.

### Bug 1 — `parseWebhookPayload` generic fallback ignored `status: resolved`

The last generic fallback in `parseWebhookPayload` hardcoded `resolved = false`, so any webhook without a `source` field, Datadog fields, or Grafana fields could **never** resolve an incident via webhook.

**Fix:** Check `status == "resolved"` in the last fallback, matching the pattern already used in the source-specific path above it.

### Bug 2 — Grafana detection matched any payload with a `title` field

The Grafana struct detection checked `gf.Title != "" || gf.RuleName != ""`. Since generic payloads also have a `title` field, they were caught by the Grafana branch and then returned `resolved = false` (because `state != "ok"`). This meant Bug 1's fix would never be reached.

**Fix:** Tighten the Grafana check to require `gf.State != "" || gf.RuleName != ""` — both are Grafana-specific fields. Real Grafana payloads always set `state`.

### Bug 3 — `extractServiceName` missed top-level `service` field

Generic webhooks send `{"service": "checkout-api", ...}` at the top level. `extractServiceName` checked `tags.service`, `commonLabels.service`, `labels.service`, etc. — but never `payload["service"]` directly. This meant `writeAffectsServiceEdge` always received an empty service name and returned early, so the `affects` edge was never written.

**Fix:** Add top-level `service` field as candidate #5 in the priority list, after Prometheus labels.

## Test Results (verified locally against live stack)

| Test | Result |
|---|---|
| Generic `status: firing` webhook → incident created | ✅ |
| Generic `status: resolved` webhook → incident resolved | ✅ (was broken) |
| Incident node promoted to `permanent` on resolve | ✅ |
| `affects` edge written from incident → service node | ✅ (was never written) |
| `similar_to` edge between two similar incidents | ✅ |
| Dedup still works | ✅ |
| Grafana webhooks (with `state` field) still detected correctly | ✅ |

Also includes: `docker-compose.yml` update to use `host.docker.internal:11434` and `phi3:mini` as Ollama defaults for macOS local dev.

### DIFF
--- a/cmd/server/graph_handlers_test.go
+++ b/cmd/server/graph_handlers_test.go
@@ -55,7 +55,7 @@ func TestHandleGetSimilarIncidents_NoGraphNodesReturnsEmptyArray(t *testing.T) {
 		t.Fatalf("CreateTeam: %v", err)
 	}
 	t.Cleanup(func() { _ = db.DeleteTeam(ctx, team.ID) })
-	incident := &store.Incident{TeamID: team.ID, Title: "Payment timeout", Severity: "critical", Status: "resolved", Source: "grafana", FiredAt: time.Now().UTC()}
+	incident := &store.Incident{TeamID: team.ID, Title: "Payment timeout", Severity: "critical", Status: "resolved", Source: "grafana", FiredAt: time.Now().UTC(), AlertPayload: []byte("{}")}
 	if err := db.CreateIncident(ctx, incident); err != nil {
 		t.Fatalf("CreateIncident: %v", err)
 	}
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -738,11 +738,18 @@ func writeForbidden(w http.ResponseWriter) {
 	_ = json.NewEncoder(w).Encode(map[string]string{"error": "forbidden"})
 }
 
-// incidentFingerprint returns a SHA-256 hex digest of team+source+title.
-// Used to deduplicate firing alerts and match resolved events to open incidents.
-func incidentFingerprint(teamID uuid.UUID, source, title string) string {
+// incidentFingerprint returns a SHA-256 hex digest of team+source+title+service.
+// The service component prevents same-title alerts from different services collapsing
+// into a single incident. Pass "" when the payload has no service field — the
+// separator is always included so the hash space stays consistent across versions.
+func incidentFingerprint(teamID uuid.UUID, source, title, service string) string {
 	h := sha256.New()
-	_, _ = fmt.Fprintf(h, "%s|%s|%s", teamID.String(), source, strings.ToLower(strings.TrimSpace(title)))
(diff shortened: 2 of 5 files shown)


fatkobra (changes requested):
Thanks — this is very close, but I’m keeping it blocked on two correctness issues.

1. **The dedup invariant is not actually enforced in the database yet.**

`CreateIncident` already treats a `23505` as a deduplicated insert, so the code is expecting a unique constraint/partial unique index to backstop the application-level lookup. But `schema.sql` only adds a non-unique index on `(team_id, fingerprint)`, so two identical webhook POSTs can still race and create duplicate incidents.

Please make this an actual unique partial index for active incidents, or add an equivalent transaction/lock-based guard.

2. **The new fingerprint still ignores the generic `service` field.**

For generic webhooks, this PR now uses the top-level `service` field for graph edges, but `incidentFingerprint(teamID, source, title)` still only keys on `source` and `title`. That means two different services that emit the same generic title will collapse into the same incident, which looks wrong for the new generic webhook path.

Please include the generic service identity in the fingerprinting path, or otherwise make sure generic alerts with the same title but different services stay distinct.

Once those two are fixed, this looks ready to go.

ntcdeveloper (commented):
Both addressed — thanks for the thorough review.

**Issue 1 — unique partial index:** it's already there, just two lines below the non-unique index you spotted. Lines 60-63 of `internal/store/schema.sql`:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_incidents_active_fingerprint_unique
    ON incidents(team_id, fingerprint)
    WHERE fingerprint IS NOT NULL
      AND status IN ('open', 'acknowledged', 'snoozed');
```

This is what the `errors.Is(err, store.ErrDuplicateIncident)` branch in `CreateIncident` relies on to catch concurrent races that slip past the application-level dedup. The `23505` handler was already wired to it. Nothing to add here.

**Issue 2 — service in fingerprint:** fixed in the last two commits.

`incidentFingerprint` now takes a `service` string as a 4th parameter — `SHA256(teamID|source|title|service)`. In `handleWebhook`, the top-level `service` field is extracted from the raw JSON payload before the fingerprint is computed, so two generic alerts with the same title but different services now produce distinct fingerprints.

Service is lowercased and trimmed to match title normalisation. Empty string when not present — consistent with other sources that don't set a top-level `service` field.

Two new tests added: `TestIncidentFingerprint_DifferentServicesDiffer` and `TestIncidentFingerprint_NormalizesService`.

fatkobra (approved):
Great work.
I approve

PR #84, opened 2026-07-09
fix: resolve mobile push notification bugs found during e2e testing

### LINKED ISSUE #83: fix: mobile push notification bugs found during end-to-end testing
## Summary

Bugs found during local end-to-end testing of the iOS mobile push notification pipeline.

### 1. Worker panic on missing incident (nil pointer dereference)
`processAlertJob` and `processResolvedIncidentJob` did not guard against `GetIncident` returning `nil, nil` (incident deleted or queue stale). Worker crashed with SIGSEGV on every stale job.

### 2. APNs push channel missing from notification rules UI
The `Channel` type in `web/app/profile/page.tsx` and `web/lib/api.ts` only included `email | sms | voice`. Users could not add a mobile push rule from the profile page.

### 3. Webhook URL shows "undefined" in team settings
`teamConfigPublic` never returned `webhook_secret`, causing the settings page to render the webhook URL as `.../undefined`. No rotate-secret endpoint was ever built, so the field is now included directly.

### 4. APNs PEM key fails to parse from env var
`parseAPNsKey` called `pem.Decode` directly on the env var value. When the key is stored with `\n` literals (standard for env vars), PEM decoding fails silently and the notifier initialises as nil.

## Also in this PR

- `docs/mobile.md` — documents the iOS app and mobile push infrastructure
- README update — mobile app listed as coming soon

ntcdeveloper:
## Summary

Bugs found and fixed during local end-to-end testing of the iOS mobile push notification pipeline. All changes are in the same commit since they were discovered together.

- **Worker panic (nil pointer)** — `processAlertJob` and `processResolvedIncidentJob` did not guard against `GetIncident` returning `nil, nil` for missing/deleted incidents. Worker crashed with SIGSEGV on every stale queue job.
- **Push channel missing from UI** — `Channel` type only had `email | sms | voice`; users could not add a mobile push rule from Profile → Notification Rules.
- **Webhook URL shows "undefined"** — `teamConfigPublic` never returned `webhook_secret`, so the settings page rendered the webhook URL as `.../undefined`.
- **APNs key fails to parse from env var** — `parseAPNsKey` called `pem.Decode` directly; keys stored with `\n` literals (standard env var format) silently failed, leaving `apnsNotifier` as nil.

## Docs

Added `docs/mobile.md` covering:
- Push relay architecture (how self-hosted customers send push without owning the APNs key)
- Relay token registration flow
- iOS app overview and minimum OS version
- Android roadmap

## Test Plan

- [x] Trigger webhook → worker processes → `✓ APNs push: 1/1 devices` in logs → push received on iPhone
- [x] Profile page → Add rule → push channel appears in dropdown
- [x] Settings page → webhook URL renders correctly (no "undefined")
- [x] Worker drains stale queue without panicking

Closes #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,6 +528,7 @@ The following scenarios have been validated end-to-end by the maintainers:
 | Grafana webhook → AI analysis → Email notification | ✅ Tested |
 | On-call escalation with voice call via Twilio | ✅ Tested |
 | Per-user notification rules (email now, voice after N min) | ✅ Tested |
+| iOS mobile push notifications via APNs | ✅ Tested |
 | Microsoft Entra SSO + group mappings | ✅ Tested |
 | GKE | 🔲 Not yet tested |
 | Datadog webhook (parser implemented) | 🔲 Not yet tested |
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -1558,6 +1558,37 @@ type teamConfigInput struct {
 	LokiEndpoint       *string  `json:"loki_endpoint"`
 }
 
+// buildTeamConfigPublic assembles the API-safe config view for a team.
+// It always fetches WebhookSecret from the teams row so both GET and PUT
+// return a consistent response — callers must not build teamConfigPublic inline.
+func (s *Server) buildTeamConfigPublic(ctx context.Context, teamID uuid.UUID, cfg *store.TeamConfig) (teamConfigPublic, error) {
+	team, err := s.db.GetTeam(ctx, teamID)
+	if err != nil {
+		return teamConfigPublic{}, fmt.Errorf("load team: %w", err)
+	}
+	if team == nil {
+		return teamConfigPublic{}, fmt.Errorf("team not found")
+	}
+
+	pub := teamConfigPublic{
+		TeamID:        teamID.String(),
+		WebhookSecret: team.WebhookSecret,
+	}
+	if cfg != nil {
+		pub.SlackWebhookURL = cfg.SlackWebhookURL
+		pub.SlackChannel = cfg.SlackChannel
+		pub.GitHubTokenSet = cfg.GitHubTokenEncrypted != nil && *cfg.GitHubTokenEncrypted != ""
+		pub.GrafanaMCPURL = cfg.GrafanaMCPURL
+		pub.GrafanaMCPTokenSet = cfg.GrafanaMCPTokenEncrypted != nil && *cfg.GrafanaMCPTokenEncrypted != ""
+		pub.PrometheusEndpoint = cfg.PrometheusEndpoint
+		pub.LokiEndpoint = cfg.LokiEndpoint
+		if cfg.GitHubRepos != nil {
+			_ = json.Unmarshal(cfg.GitHubRepos, &pub.GitHubRepos)
+		}
+	}
+	return pub, nil
+}
+
 func (s *Server) handleGetTeamConfig(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	teamID, err := uuid.Parse(vars["teamId"])
@@ -1576,21 +1607,10 @@ func (s *Server) handleGetTeamConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	pub := teamConfigPublic{TeamID: teamID.String()}
-	if team, err := s.db.GetTeam(r.Context(), teamID); err == nil && team != nil {
-		pub.WebhookSecret = team.WebhookSecret
-	}
-	if cfg != nil {
-		pub.SlackWebhookURL = cfg.SlackWebhookURL
-		pub.SlackChannel = cfg.SlackChannel
-		pub.GitHubTokenSet = cfg.GitHubTokenEncrypted != nil && *cfg.GitHubTokenEncrypted != ""
-		pub.GrafanaMCPURL = cfg.GrafanaMCPURL
-		pub.GrafanaMCPTokenSet = cfg.GrafanaMCPTokenEncrypted != nil && *cfg.GrafanaMCPTokenEncrypted != ""
-		pub.PrometheusEndpoint = cfg.PrometheusEndpoint
-		pub.LokiEndpoint = cfg.LokiEndpoint
-		if cfg.GitHubRepos != nil {
-			_ = json.Unmarshal(cfg.GitHubRepos, &pub.GitHubRepos)
-		}
+	pub, err := s.buildTeamConfigPublic(r.Context(), teamID, cfg)
+	if err != nil {
+		http.Error(w, "failed to load team", http.StatusInternalServerError)
+		return
 	}
 
 	w.Header().Set("Content-Type", "application/json")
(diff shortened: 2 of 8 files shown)


fatkobra (changes requested):
****1.** webhook_secret is fixed for GET, but still missing/empty after config saves**

The PR adds WebhookSecret to teamConfigPublic and populates it in handleGetTeamConfig, which fixes the initial Settings page load.

However, handleUpsertTeamConfig still returns a teamConfigPublic without setting WebhookSecret. The Settings page calls setConfig(updated) after both saveDataSources and saveNotifications, and the displayed webhook URL is derived from config.team_id and config.webhook_secret. That means the Settings page can regress to a broken webhook URL immediately after a successful save, because the PUT response replaces the previously loaded config with one whose webhook_secret is empty.

Suggested fix: centralize the public config response builder and use it from both GET and PUT:

````
func (s *Server) buildTeamConfigPublic(ctx context.Context, teamID uuid.UUID, cfg *store.TeamConfig) (teamConfigPublic, error) {
    team, err := s.db.GetTeam(ctx, teamID)
    if err != nil {
        return teamConfigPublic{}, fmt.Errorf("load team: %w", err)
    }
    if team == nil {
        return teamConfigPublic{}, pgx.ErrNoRows
    }

    pub := teamConfigPublic{
        TeamID:        teamID.String(),
        WebhookSecret: team.WebhookSecret,
    }

    if cfg != nil {
        pub.SlackWebhookURL = cfg.SlackWebhookURL
        pub.SlackChannel = cfg.SlackChannel
        pub.GitHubTokenSet = cfg.GitHubTokenEncrypted != nil && *cfg.GitHubTokenEncrypted != ""
        pub.GrafanaMCPURL = cfg.GrafanaMCPURL
        pub.GrafanaMCPTokenSet = cfg.GrafanaMCPTokenEncrypted != nil && *cfg.GrafanaMCPTokenEncrypted != ""
        pub.PrometheusEndpoint = cfg.PrometheusEndpoint
        pub.LokiEndpoint = cfg.LokiEndpoint
        if cfg.GitHubRepos != nil {
            _ = json.Unmarshal(cfg.GitHubRepos, &pub.GitHubRepos)
        }
    }

    return pub, nil
}
````
Then call this helper from both handleGetTeamConfig and handleUpsertTeamConfig. I would also avoid silently swallowing GetTeam errors in GET; returning an empty secret is worse than returning a clear 500/404.

**2. docs/mobile.md describes relay config that does not exist in this PR**

The new docs say self-hosted deployments should configure a hosted relay using Helm values like:
````
push:
  relayURL: https://push.wachd.io
  relayToken: wpr_xxxxxxxxxxxx
````
and environment variables like WACHD_PUSH_RELAY_URL / WACHD_PUSH_RELAY_TOKEN.

But the worker still constructs direct APNs/FCM notifiers: notify.NewAPNsNotifier() and notify.NewFCMNotifier(). The APNs notifier requires APNS_KEY_ID, APNS_TEAM_ID, APNS_BUNDLE_ID, and APNS_PRIVATE_KEY, and there is no WACHD_PUSH_RELAY reference in the worker or APNs notifier.

The Helm chart also exposes direct APNs/FCM values under config.notifications.apns and config.notifications.fcm, not push.relayURL / push.relayToken.

This is user-impacting: a self-hosted customer following the new docs will configure values that the application never reads, and push will remain disabled.

Suggested fix: either implement the relay notifier in this PR, or change docs/mobile.md to match the current direct APNs/FCM implementation. If relay is the intended production path, I would add a RelayNotifier and make worker selection explicit:

````
relayURL := os.Getenv("WACHD_PUSH_RELAY_URL")
relayToken := os.Getenv("WACHD_PUSH_RELAY_TOKEN")

if relayURL != "" && relayToken != "" {
    pushNotifier = notify.NewRelayNotifier(relayURL, relayToken)
} else {
    apnsNotifier = notify.NewAPNsNotifier()
    fcmNotifier = notify.NewFCMNotifier()
}
````
Then add matching Helm values/templates for WACHD_PUSH_RELAY_URL and WACHD_PUSH_RELAY_TOKEN.

fatkobra:
### Smaller but worthwhile follow-ups
**3. Add regression tests for the exact bugs fixed here**

There are already APNs/FCM tests, but I do not see a test that passes a valid PEM with literal \n sequences through APNS_PRIVATE_KEY. Since this was one of the concrete e2e failures, it deserves a targeted regression test.

Example:
````
func TestNewAPNsNotifier_AcceptsEscapedNewlinesInPrivateKey(t *testing.T) {
    keyPEM := generateTestECKeyPEM(t)
    escaped := strings.ReplaceAll(keyPEM, "\n", `\n`)

    t.Setenv("APNS_KEY_ID", "TESTKEY0001")
    t.Setenv("APNS_TEAM_ID", "TESTTEAM001")
    t.Setenv("APNS_BUNDLE_ID", "io.wachd.test")
    t.Setenv("APNS_PRIVATE_KEY", escaped)

    if n := NewAPNsNotifier(); n == nil {
        t.Fatal("expected APNs notifier with escaped-newline PEM")
    }
}
````
I would also add tests for both stale queue paths: processAlertJob and processResolvedIncidentJob should return nil when GetIncident returns nil, nil. The PR adds those guards in both places.


**4. APNs parse failures are still silent**

This PR fixes the common env-var newline case, but NewAPNsNotifier() still returns nil on any parse failure without logging why. The issue summary explicitly mentions that the old behavior “failed silently,” and that part remains true for malformed keys, wrong key types, or copied credentials with bad whitespace.

A small improvement would be to return (*APNsNotifier, error) or add a worker-side config validation/log path so operators see:
````
APNs push notifier disabled: failed to decode PEM block from APNS_PRIVATE_KEY
````
That would make future e2e/debugging much easier.

**5. UX wording for push rules is awkward**

The Profile page renders every rule as “Notify me via X at {session.email}.” That works for email, but reads oddly for mobile push: “Notify me via mobile push at user@example.com.” The PR adds push to the dropdown and labels it “mobile push,” which is good; I would special-case the target text for push/SMS/voice.

For example:

````
const target =
  rule.channel === 'email'
    ? `at ${session?.email ?? 'your email'}`
    : rule.channel === 'push'
      ? 'on registered mobile devices'
      : '';
````

ntcdeveloper (commented):
Thanks for the thorough review.

**Issue 1 — webhook_secret missing after PUT:** Fixed. Extracted `buildTeamConfigPublic` helper that fetches the team row (for the secret) and assembles the full public view. Both `handleGetTeamConfig` and `handleUpsertTeamConfig` now call it. `GetTeam` errors are also no longer swallowed — they return a 500 instead of an empty secret.

**Issue 2 — docs describe non-existent relay config:** Fixed. Rewrote `docs/mobile.md` to describe the actual direct APNs/FCM env var configuration. The relay architecture is the right long-term design but implementing `RelayNotifier` is a separate piece of work — will be added in a follow-up PR.

**Issue 3 — regression test for escaped-newline PEM:** Added `TestNewAPNsNotifier_AcceptsEscapedNewlinesInPrivateKey` in `apns_fcm_test.go`. Passes.

For the worker nil guard tests (`processAlertJob`/`processResolvedIncidentJob`): `Worker.db` is a concrete `*store.DB`, not an interface, so these can't be unit tested without a real DB connection. The existing `store_test.go` already verifies that `GetIncident` returns `nil, nil` for a missing incident ID. Adding an interface to enable mocking here is a larger refactor — happy to do it as a follow-up if you think it's worth it.

**Issue 4 — silent APNs parse failures:** Fixed. `NewAPNsNotifier` now logs the reason before returning nil (e.g. `APNs push notifier disabled: failed to decode PEM block from APNS_PRIVATE_KEY`).

**Issue 5 — push rule UX wording:** Fixed. Push channel now shows "on registered mobile devices" instead of "at {email}".

fatkobra:
Looking goo.
Approved.

